PR #8, opened 2024-02-22
Fix todos

Marenz:
- Update versions
- Fix cookiecutter TODOS


.github/labeler.yml
```diff
@@ -61,3 +61,8 @@
       - CODEOWNERS
       - MANIFEST.in
       - noxfile.py
+
+"part:actor":
+  - changed-files:
+    - any-glob-to-any-file:
+      - "src/frequenz/actor/dispatch/**"
```
llucax: With this repo scheme, I would put the actor in `frequenz/dispatch/actor` instead, as the python package is using the `frequenz.dispatch` namespace.

mkdocs.yml
```diff
@@ -116,12 +112,12 @@ plugins:
             show_source: true
             signature_crossrefs: true
           import:
-            # TODO(cookiecutter): You might want to add other external references here
             # See https://mkdocstrings.github.io/python/usage/#import for details
             - https://docs.python.org/3/objects.inv
             - https://frequenz-floss.github.io/frequenz-channels-python/v0.16/objects.inv
             - https://frequenz-floss.github.io/frequenz-sdk-python/v0.25/objects.inv
```
llucax, line 118: Side note: Oh, dammit, we need to depend on the SDK for the actor, but the SDK will depend on the actor too eventually, we need to address https://github.com/frequenz-floss/frequenz-sdk-python/issues/851 sooner than I expected.

.github/labeler.yml
Marenz: Updated

llucax: Maybe not? :P

Marenz: On second glance, this involved WAY more files than I anticipated.

Marenz: Hmm but the client has the structure/namespace `frequenz.client.dispatch`

llucax: > On second glance, this involved WAY more files than I anticipated.

Why?

> Hmm but the client has the structure/namespace `frequenz.client.dispatch`

Yes, and it is in a repo called `frequenz-client-dispatch`. I mean, we can just name this repo `frequenz-actor-dispatch` and have only the actor as we planned originally, I just thought it might have been a good opportunity to try this other scheme, by mixing the actor and high-level interface in one repo, and keeping the actor as an implementation detail, as we do in the SDK, where users never need to deal with the actual actors used behind the scenes, and just use higher-level functions and objects that interact with the actors for them.

llucax: Oh, I see the diff now, so you actually used the `actor` template to create this repo, not the `lib` template, right?

.cookiecutter-replay.json
```diff
@@ -10,8 +10,8 @@
     "license": "MIT",
     "author_name": "Frequenz Energy-as-a-Service GmbH",
     "author_email": "floss@frequenz.com",
-    "python_package": "frequenz.actor.dispatch",
-    "pypi_package_name": "frequenz-actor-dispatch",
+    "python_package": "frequenz.dispatch.actor",
+    "pypi_package_name": "frequenz-dispatch-actor",
```
llucax:
```suggestion
    "python_package": "frequenz.dispatch",
    "pypi_package_name": "frequenz-dispatch",
```

CONTRIBUTING.md
```diff
@@ -141,7 +141,7 @@ These are the steps to create a new release:
    a [GitHub
    Release](https://github.com/frequenz-floss/frequenz-dispatch-python/releases),
    and upload a new package to
-   [PyPI](https://pypi.org/project/frequenz-actor-dispatch/)
+   [PyPI](https://pypi.org/project/frequenz-dispatch-actor/)
```
llucax:
```suggestion
   [PyPI](https://pypi.org/project/frequenz-dispatch/)
```

README.md
```diff
@@ -1,14 +1,37 @@
 # Dispatch Highlevel Interface
 
 [![Build Status](https://github.com/frequenz-floss/frequenz-dispatch-python/actions/workflows/ci.yaml/badge.svg)](https://github.com/frequenz-floss/frequenz-dispatch-python/actions/workflows/ci.yaml)
-[![PyPI Package](https://img.shields.io/pypi/v/frequenz-actor-dispatch)](https://pypi.org/project/frequenz-actor-dispatch/)
+[![PyPI Package](https://img.shields.io/pypi/v/frequenz-dispatch-actor)](https://pypi.org/project/frequenz-dispatch-actor/)
```
llucax:
```suggestion
[![PyPI Package](https://img.shields.io/pypi/v/frequenz-dispatch)](https://pypi.org/project/frequenz-dispatch/)
```

pyproject.toml
```diff
@@ -5,12 +5,12 @@
...
   "setuptools == 68.1.0",
   "setuptools_scm[toml] == 7.1.0",
-  "frequenz-repo-config[actor] == 0.8.0",
+  "frequenz-repo-config[actor] == 0.9.0",
 ]
 build-backend = "setuptools.build_meta"
 
 [project]
-name = "frequenz-actor-dispatch"
+name = "frequenz-dispatch-actor"
```
llucax:
```suggestion
name = "frequenz-dispatch"
```

pyproject.toml
```diff
@@ -43,48 +43,48 @@ email = "floss@frequenz.com"
...
   "mkdocstrings[python] == 0.23.0",
-  "frequenz-repo-config[actor] == 0.8.0",
+  "frequenz-repo-config[actor] == 0.9.0",
 ]
 dev-mypy = [
   "mypy == 1.5.1",
   "types-Markdown == 3.4.2.10",
   # For checking the noxfile, docs/ script, and tests
-  "frequenz-actor-dispatch[dev-mkdocs,dev-noxfile,dev-pytest]",
+  "frequenz-dispatch-actor[dev-mkdocs,dev-noxfile,dev-pytest]",
```
llucax:
```suggestion
  "frequenz-dispatch[dev-mkdocs,dev-noxfile,dev-pytest]",
```

pyproject.toml
```diff
@@ -43,48 +43,48 @@ email = "floss@frequenz.com"
...
 ]
 dev-noxfile = [
   "nox == 2023.4.22",
-  "frequenz-repo-config[actor] == 0.8.0",
+  "frequenz-repo-config[actor] == 0.9.0",
```
llucax:
```suggestion
  "frequenz-repo-config[lib] == 0.9.0",
```

pyproject.toml
```diff
@@ -43,48 +43,48 @@ email = "floss@frequenz.com"
...
+  "black == 24.2.0",
   "Markdown==3.4.4",
   "mike == 2.0.0",
   "mkdocs-gen-files == 0.5.0",
   "mkdocs-literate-nav == 0.6.1",
   "mkdocs-macros-plugin == 1.0.4",
   "mkdocs-material == 9.3.1",
   "mkdocstrings[python] == 0.23.0",
-  "frequenz-repo-config[actor] == 0.8.0",
+  "frequenz-repo-config[actor] == 0.9.0",
```
llucax:
```suggestion
  "frequenz-repo-config[lib] == 0.9.0",
```

pyproject.toml
```diff
@@ -5,12 +5,12 @@
 requires = [
   "setuptools == 68.1.0",
   "setuptools_scm[toml] == 7.1.0",
-  "frequenz-repo-config[actor] == 0.8.0",
+  "frequenz-repo-config[actor] == 0.9.0",
```
llucax:
```suggestion
  "frequenz-repo-config[lib] == 0.9.0",
```

pyproject.toml
```diff
@@ -43,48 +43,48 @@ email = "floss@frequenz.com"
...
 dev-noxfile = [
   "nox == 2023.4.22",
-  "frequenz-repo-config[actor] == 0.8.0",
+  "frequenz-repo-config[actor] == 0.9.0",
 ]
 dev-pylint = [
   "pylint == 3.0.2",
   # For checking the noxfile, docs/ script, and tests
-  "frequenz-actor-dispatch[dev-mkdocs,dev-noxfile,dev-pytest]",
+  "frequenz-dispatch-actor[dev-mkdocs,dev-noxfile,dev-pytest]",
```
llucax:
```suggestion
  "frequenz-dispatch[dev-mkdocs,dev-noxfile,dev-pytest]",
```

pyproject.toml
```diff
@@ -43,48 +43,48 @@ email = "floss@frequenz.com"
...
   "pytest == 8.0.0",
-  "frequenz-repo-config[extra-lint-examples] == 0.8.0",
+  "frequenz-repo-config[extra-lint-examples] == 0.9.0",
   "pytest-mock == 3.11.1",
   "pytest-asyncio == 0.21.1",
   "async-solipsism == 0.5",
 ]
 dev = [
-  "frequenz-actor-dispatch[dev-mkdocs,dev-flake8,dev-formatting,dev-mkdocs,dev-mypy,dev-noxfile,dev-pylint,dev-pytest]",
+  "frequenz-dispatch-actor[dev-mkdocs,dev-flake8,dev-formatting,dev-mkdocs,dev-mypy,dev-noxfile,dev-pylint,dev-pytest]",
```
llucax:
```suggestion
  "frequenz-dispatch[dev-mkdocs,dev-flake8,dev-formatting,dev-mkdocs,dev-mypy,dev-noxfile,dev-pylint,dev-pytest]",
```

pyproject.toml
```diff
@@ -159,7 +159,7 @@ namespace_packages = true
 # used but getting the original ignored error when removing the type: ignore.
 # See for example: https://github.com/python/mypy/issues/2960
 #no_incremental = true
-packages = ["frequenz.actor.dispatch"]
+packages = ["frequenz.dispatch.actor"]
```
llucax:
```suggestion
packages = ["frequenz.dispatch"]
```

src/frequenz/dispatch/actor/__init__.py
llucax, line 1: `src/frequenz/dispatch/__init__.py` should also exist, the end user should probably do something like:

```py
from frequenz.dispatch import Dispatcher

async def run():
    dispatcher = Dispatcher(API_CONNECTION_INFO)
    dispatcher.start()  # this will fire the actor
    dispatch_arrived = dispatcher.new_dispatches().new_receiver()
    dispatch_ready = dispatcher.ready_dispatches().new_receiver()
    # ...
```

I mean, in this case maybe the actor could be the same as the `Dispatcher` object, but so far we never put any methods in actors to make it clear the communication to actors should be done via channels instead. Maybe we should think about this possibility instead of having this in 2 layers, but this is a broader question, affecting other APIs too I guess.

Then the SDK will take care of create a global `Dispatcher`, so users using the SDK should do something more along the lines:

```py
from frequenz import sdk

sdk.initialize(CONFIG)

async def run():
    dispatch_arrived = sdk.dispatcher.new_dispatches().new_receiver()
    dispatch_ready = sdk.dispatcher.ready_dispatches().new_receiver()
    # ...
```

src/frequenz/dispatch/actor/conftest.py
llucax, line 1: Same here, this should stay in `src/frequenz/dispatch/conftest.py`, etc.

llucax (commented):
OK, it seems there has been quite some confusion about the repo/python package name and general structure. The idea of calling this repo `frequenz-dispatch-python` instead of `frequenz-actor-dispatch-python` was to provide both the *high-level* interface and the *actor* in one repo (and use the normal `lib` repo config template).

[image](https://github.com/frequenz-floss/frequenz-sdk-python/wiki/APIs-stack-and-repositories-structure)

Sorry about the confusion and extra work, we can of course talk about it if you don't agree and go with a different approach, it's that because you specifically pointed out that having so many repos was confusing I saw this as one place where we might reduce the number of repos, and thought it could help make things more clear instead of more confusion :(

.github/labeler.yml
Marenz: Yes, I used `actor`

>>On second glance, this involved WAY more files than I anticipated.
>
>Why?

Because the name/path `frequenz-actor-dispatch` was all over the code, dependencies, tests, etc

> I just thought it might have been a good opportunity to try this other scheme

Ah, alright. Sure, we can try.

.github/workflows/ci.yaml
llucax, line 1: The changes to this file are in the wrong commit.

pyproject.toml
```diff
@@ -5,12 +5,12 @@
 requires = [
   "setuptools == 68.1.0",
   "setuptools_scm[toml] == 7.1.0",
-  "frequenz-repo-config[actor] == 0.8.0",
+  "frequenz-repo-config[lib] == 0.9.0",
```
llucax, line 8: You could bump directly to v0.9.1 here, but dependabot will take care of it if you don't.

llucax (approved):
It would be good to fix at leas the version bump in the wrong commit, but even that is not that important, so I'll approve anyway.

.github/workflows/ci.yaml
Marenz: fixed. No re-approve required.

PR #9, opened 2024-02-28
Initial structure

Marenz:
- Prepare dependencies for dispatch
- Fix labeler path
- Add initial structure


src/frequenz/dispatch/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+
+    ```python
+    from frequenz.dispatch import Dispatcher
+
+    async def run():
+        dispatcher = Dispatcher(API_CONNECTION_INFO)
+        dispatcher.start()  # this will start the actor
+        dispatch_arrived = dispatcher.new_dispatches().new_receiver()
+        dispatch_ready = dispatcher.ready_dispatches().new_receiver()
+    ```
```
llucax:
```suggestion
    Example:
        ```python
        from frequenz.dispatch import Dispatcher

        async def run():
            dispatcher = Dispatcher(API_CONNECTION_INFO)
            dispatcher.start()  # this will start the actor
            dispatch_arrived = dispatcher.new_dispatches().new_receiver()
            dispatch_ready = dispatcher.ready_dispatches().new_receiver()
        ```
```

shsms:
Could you describe the motivation for this change?  There's `from frequenz import dispatch` now?  Aren't we only having namespace packages under frequenz like `frequenz.{actor, client, api, etc}` and putting code under that?

Or is this going to be a completely independent thing like `frequenz.channels`?

Marenz:
The structure is a kind of experiment that luca suggested to me to try. 
Basically, as the user shouldn't interact directly with the actor, the high-level interface should just directly offer the dispatch arrived/is ready channels like shown in https://github.com/frequenz-io/frequenz-apps-draft/blob/main/dispatch_exec/dispatch.py

src/frequenz/dispatch/actor.py
```diff
@@ -0,0 +1,199 @@
...
+        self._scheduled: dict[int, asyncio.Task[None]] = {}
+        self._microgrid_id = microgrid_id
+        self._updated_dispatch_sender = updated_dispatch_sender
+        self._ready_dispatch_sender = ready_dispatch_sender
+
+    async def _run(self) -> None:
+        """Run the actor."""
+        try:
+            while True:
+                # for _ in range(30):
```
shsms: I guess this can go away.

src/frequenz/dispatch/actor.py
```diff
@@ -0,0 +1,199 @@
...
+                await self._fetch()
+                await asyncio.sleep(timedelta(hours=5).total_seconds())
```
shsms: maybe you want to take the 5 hours from a constructor argument with a default value.  That way you can give it a name and a description, so people would know why 5 hours.

src/frequenz/dispatch/actor.py
```diff
@@ -0,0 +1,199 @@
...
+                task.cancel()
+            raise
+
+    async def _fetch(self) -> None:
+        """Fetch all relevant dispatches."""
+        old_dispatches = self._dispatches
+        self._dispatches = {}
+
+        try:
+            logging.info("Fetching dispatches for microgrid %s", self._microgrid_id)
```
shsms: I think we should always have `_logger = logging.getLogger(__name__)` at the top of the file and use `_logger.info`, etc. everywhere.

With that, we'll be able to do filter levels like `logging.getLogger("frequenz.actor").setLevel(logging.DEBUG)`, and I think we should support that in all our packages.

shsms: Maybe we should look into making a pylint plugin for that.

src/frequenz/dispatch/actor.py
```diff
@@ -0,0 +1,199 @@
...
+            await asyncio.sleep(
+                (
+                    next_time - datetime.now().replace(tzinfo=timezone.utc)
+                ).total_seconds()
+            )
+            logging.info("Dispatch ready: %s", dispatch)
+            await self._ready_dispatch_sender.send(dispatch)
+
+    @classmethod
+    def calculate_next_run(cls, dispatch: Dispatch, _from: datetime) -> datetime | None:
```
shsms, line 233: why does this have to be a public class method?  from what I can see, it will be used only from within the `Dispatcher`, and only from contexts where a `self` is available.

src/frequenz/dispatch/actor.py
```diff
@@ -0,0 +1,199 @@
...
+            logging.info("Dispatch %s scheduled for %s", dispatch.id, next_time)
+            logging.info(
+                "Sleeping for %s",
+                next_time - datetime.now().replace(tzinfo=timezone.utc),
+            )
+            await asyncio.sleep(
+                (
+                    next_time - datetime.now().replace(tzinfo=timezone.utc)
+                ).total_seconds()
+            )
```
shsms: maybe we can reuse the `now()` values, because it is a separate syscall each time.

shsms: also,  isn't this the same as `datetime.now(tz=timezone.utc)`?  Or does `replace` have a different behaviour?

src/frequenz/dispatch/actor.py
```diff
@@ -0,0 +1,199 @@
...
+            dispatch: The dispatch to schedule.
+        """
+        while True:
+            next_time = self.calculate_next_run(
+                dispatch, datetime.now().replace(tzinfo=timezone.utc)
+            )
+            if next_time is None:
+                logging.info("Dispatch finished: %s", dispatch)
+                self._scheduled.pop(dispatch.id)
+                return
```
shsms: maybe more readable with:

```python
while next_time := self.calculate_next_run(dispatch, datetime.now(tz=timezone.utc)):
    ...

# and after the loop
logging.info("Dispatch finished: %s", dispatch)
self._scheduled.pop(dispatch.id)
```

shsms:
> The structure is a kind of experiment that luca suggested to me to try.
> Basically, as the user shouldn't interact directly with the actor, the high-level interface should just directly offer the dispatch arrived/is ready channels like shown in https://github.com/frequenz-io/frequenz-apps-draft/blob/main/dispatch_exec/dispatch.py

Then maybe we keep `frequenz.actor.dispatch` as it is, and add `frequenz.dispatch` separately, such that `frequenz.dispatch` can eventually go into the SDK?

Either way, the reason for the change is not obvious at all, so it would be worth the effort to explain it in the commit message and in the PR description.

src/frequenz/dispatch/actor.py
Marenz: I meant to use the constant (which is also missing docs) `_MAX_AHEAD_SCHEDULE`

src/frequenz/dispatch/actor.py
Marenz: It doesn't have to be, but as a user who wants to create new dispatches and investigate existing ones, I figured it's very useful to know when the next time is a dispatch would run.

src/frequenz/dispatch/actor.py
Marenz: it's pretty much the same

src/frequenz/dispatch/actor.py
Marenz: then I can't reuse `now` easily with your suggestion, but I modified it a bit..

```python
        def next_run_info() -> tuple[datetime, datetime] | None:
            now = datetime.now(tz=timezone.utc)
            next_run = self.calculate_next_run(dispatch, now)

            if next_run is None:
                return None

            return now, next_run

        while pair := next_run_info():
            now, next_time = pair

            if next_time - now > _MAX_AHEAD_SCHEDULE:
                await asyncio.sleep(_MAX_AHEAD_SCHEDULE.total_seconds())
                continue

            _logger.info("Dispatch %s scheduled for %s", dispatch.id, next_time)
            _logger.info("Sleeping for %s", next_time - now)
            await asyncio.sleep((next_time - now).total_seconds())

            _logger.info("Dispatch ready: %s", dispatch)
            await self._ready_dispatch_sender.send(dispatch)

        _logger.info("Dispatch finished: %s", dispatch)
        self._scheduled.pop(dispatch.id)

```

Marenz:
I don't see it much as "change of structure", given that previously there was basically nothing except empty or dummy files  ;)

Anyway, I found what I was missing. The structure change is an attempt to test the rewrite of the structure according to https://github.com/frequenz-floss/frequenz-sdk-python/wiki/APIs-stack-and-repositories-structure#completely-rewrite-of-the-structure

Marenz:
* Added link to structure rewrite description in commit
* implemented proper scheduling with recurrence based on `python-dateutils`
* can use more tests, but that can be added later (test suggestions welcome, part from the create/delete/update tests that are there/planned)

README.md
```diff
@@ -15,22 +15,30 @@ It provides two channels for clients:
...
+                match selected.value:
+                    case Created(dispatch):
+                        log.info("New dispatch arrived %s", dispatch.created)
+                        ...
+                    case Updated(dispatch):
+                        log.info("Dispatch updated %s", dispatch.updated)
+                        ...
+                    case Deleted(dispatch):
+                        log.info("Dispatch deleted %s", dispatch.deleted)
+                        ...
```
shsms: I think you don't need the `.created`, etc. anymore.

shsms:
Looks good to me.

src/frequenz/dispatch/__init__.py
```diff
@@ -0,0 +1,71 @@
...
+from frequenz.dispatch.actor import DispatchActor, DispatchEvent
+
+__all__ = ["Dispatcher"]
+
+
+class Dispatcher:
+    """A highlevel interface for the dispatch API.
+
+    This class provides a highlevel interface to the dispatch API. It
+    allows to receive new dispatches and ready dispatches.
```
stefan-brus-frequenz, line 26: Suggestion:

```
This class provides a highlevel interface to the dispatch API. It 
allows for receiving of new dispatches, and dispatches labelled as 'ready'.
```

I would also suggest adding a small blurb about the difference between the two.

src/frequenz/dispatch/actor.py
```diff
@@ -0,0 +1,275 @@
...
+            await self._updated_dispatch_sender.send(Deleted(dispatch=dispatch))
+            if task := self._scheduled.pop(dispatch.id, None):
+                task.cancel()
+
+    def _update_dispatch_schedule(
+        self, dispatch: Dispatch, old_dispatch: Dispatch | None
+    ) -> None:
+        """Update the schedule for a dispatch.
+
+        Schedules, reschedules or cancels the dispatch based on the start_time
```
stefan-brus-frequenz, line 174: Could you clarify how this happens in this doc?

For example, "- If the start time of the new dispatch is different from the old one, the old one is cancelled."

stefan-brus-frequenz (dismissed):
Added a couple of optional comments about the docs, code overall LGTM, but my python knowledge is limited.

stefan-brus-frequenz (approved):

pyproject.toml
```diff
@@ -64,11 +69,13 @@ dev-mkdocs = [
 dev-mypy = [
   "mypy == 1.8.0",
   "types-Markdown == 3.5.0.20240129",
+  "types-python-dateutil==2.8.19.20240311",
   # For checking the noxfile, docs/ script, and tests
   "frequenz-dispatch[dev-mkdocs,dev-noxfile,dev-pytest]",
 ]
 dev-noxfile = [
-  "nox == 2023.4.22",
+  "uv == 0.1.14",
```
llucax, line 77: :scream:

pyproject.toml
```diff
@@ -26,21 +25,27 @@ classifiers = [
...
 dependencies = [
+  "python-dateutil >= 2.8.2, < 3.0",
   "typing-extensions == 4.10.0",
   # Make sure to update the version for cross-referencing also in the
   # mkdocs.yml file when changing the version here (look for the config key
   # plugins.mkdocstrings.handlers.python.import)
-  "frequenz-sdk == 0.25.2",
+  "frequenz-sdk == v1.0.0-rc5",
+  "frequenz-channels == 1.0.0b2",
+  "frequenz-dispatch-client @ git+https://github.com/frequenz-floss/frequenz-client-dispatch-python.git@00dcb3c",
```
llucax, line 36: Oh, damn, this python package name is wrong, it should be `frequenz-client-dispatch`).

* https://github.com/frequenz-floss/frequenz-client-dispatch-python/pull/27

pyproject.toml
```diff
@@ -26,21 +25,27 @@ classifiers = [
...
+  "frequenz-api-dispatch @ git+https://github.com/frequenz-floss/frequenz-api-dispatch.git@223315c",
+#  "frequenz-api-dispatch >= 0.13.0, < 0.14",
+  "frequenz-client-base >= 0.2.1, < 0.3.0",
+  # Directly use unreleased commit until the first release
+  "frequenz-client-common @ git+https://github.com/frequenz-floss/frequenz-client-common-python.git@5e6e7b7",
```
llucax, line 41: These 3 are here because of `uv`? Or are they really used directly by this project? And if so, why, intuitively this project should only depend on `client-dispatch`.

src/frequenz/dispatch/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+
+"""A highlevel interface for the dispatch API."""
+
+import grpc.aio
+from frequenz.channels import Broadcast, Receiver
+from frequenz.client.dispatch.types import Dispatch
+
+from frequenz.dispatch.actor import DispatchActor, DispatchEvent
+
+__all__ = ["Dispatcher"]
```
llucax, line 12: We need to export more stuff here, all that's returned by the public interface too, like `DispatchEvent` and all the stuff that's exposed by it.

src/frequenz/dispatch/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class Dispatcher:
```
llucax, line 15: Since we normally don't put actual code in `__init__.py`, I suggest moving this to `_dispatcher.py`.

src/frequenz/dispatch/actor.py
```diff
@@ -0,0 +1,279 @@
...
+_MAX_AHEAD_SCHEDULE = timedelta(hours=5)
+"""The maximum time ahead to schedule a dispatch.
+
+We don't want to schedule dispatches too far ahead,
+as they could start drifting if the delay is too long.
+
+This also prevents us from scheduling too many dispatches at once.
+
+The exact value is not important, but should be a few hours and not more than a day.
+"""
```
llucax, line 28: I didn't understand which issue this is trying to solve, maybe we need a more detailed explanation of the issue and how this solves it.

src/frequenz/dispatch/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+            svc_addr,
+            self._updated_channel.new_sender(),
+            self._ready_channel.new_sender(),
+        )
+
+    async def start(self) -> None:
+        """Start the actor."""
+        self._actor.start()
+
+    def updated_dispatches(self) -> Receiver[DispatchEvent]:
```
llucax, line 64: Now that this handles create and delete too, maybe `updated` is not the most clear choice. When I described the issue, ChatGPT suggested `dispatch_lifecycle_events` and `ready_for_execution_dispatches` but it seems a bit too long. I wonder if we really need to add `dispatch` to the name, as it should be pretty obvious in this object's context.

Maybe just `dispatcher.lifecycle_events` and `dispatcher.ready_to_execute`?

Another approach would be to unify both channels and just have `dispatcher.dispatches`, but I guess many users will just want to just subscribe to the ready dispatches and all the rest will be (a considerable amount of) noise.

src/frequenz/dispatch/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+        """Return new, updated or deleted dispatches receiver.
+
+        Returns:
+            A new receiver for new dispatches.
+        """
+        return self._updated_channel.new_receiver()
```
llucax, line 70: For the SDK we decided to call stuff creating new objects more clearly, as people might think that a receiver is stored internally. I think we should do the same here, so it should be something like: `dispatcher.lifecycle_events.new_receiver()` and `dispatcher.ready_to_execute.new_receiver()`? (following the suggestion above)

We can just return the internal channel if we change the return type to a [`ReceiverFetcher`](https://github.com/frequenz-floss/frequenz-sdk-python/blob/1a3dce7623684120c999ff23b86fa984a4890471/src/frequenz/sdk/_internal/_channels.py#L15-L27) protocol (we can just copy&paste the protocol, it is really super lightweight.

So the implementation can simply be:

```py
    @property
    def lifecycle_events(self) -> ReceiverFetcher[DispatchEvent]:
        return self._updated_channel
```

src/frequenz/dispatch/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def ready_dispatches(self) -> Receiver[Dispatch]:
```
llucax, line 72: Same here.

src/frequenz/dispatch/actor.py
```diff
@@ -0,0 +1,279 @@
...
+    Weekday.THURSDAY: rrule.TH,
+    Weekday.FRIDAY: rrule.FR,
+    Weekday.SATURDAY: rrule.SA,
+    Weekday.SUNDAY: rrule.SU,
+}
+"""To map from our Weekday enum to the dateutil library enum."""
+
+
+@dataclass(frozen=True)
+class _DispatchEventBase:
```
llucax, line 59: This should be public (`DispatchEventBase`) if public classes inherit from it. But I think, being just a super thin wrapper, we should just not inherit from a common base, which also could lead to user mistakes if they use the base class in a `match` statement.

src/frequenz/dispatch/actor.py
```diff
@@ -0,0 +1,279 @@
...
+
+DispatchEvent = Created | Updated | Deleted
+"""Type that is sent over the channel for dispatch updates.
+
+This type is used to send dispatches that were created, updated or deleted
+over the channel.
+"""
+
+
+class DispatchActor(Actor):
```
llucax, line 88: Technically this should be named `DispatchingActor` according to the naming convention :grimacing:

src/frequenz/dispatch/actor.py
```diff
@@ -0,0 +1,279 @@
...
+        self._updated_dispatch_sender = updated_dispatch_sender
+        self._ready_dispatch_sender = ready_dispatch_sender
+        self._poll_interval = poll_interval
+
+    async def _run(self) -> None:
+        """Run the actor."""
+        try:
+            while True:
+                await self._fetch()
+                await asyncio.sleep(self._poll_interval.total_seconds())
```
llucax, line 132: We should use a `Timer` here to properly deal with drift, even when unlikely (it will make the code simpler too):
`async for drift in self._poll_timer: await self._fetch()` (we could also print the drift or something).

src/frequenz/dispatch/actor.py
```diff
@@ -0,0 +1,279 @@
...
+        except asyncio.CancelledError:
+            for task in self._scheduled.values():
+                task.cancel()
+            raise
```
llucax, line 136: We need https://github.com/frequenz-floss/frequenz-sdk-python/issues/819!

llucax, line 136: It took me a while to think about the different scenarios to conclude calling cancel only when `CancelledError` is raised. Just to confirm my thinking is correct, you do this because if any other unexpected exception is raised and `_run()` is aborted, the auto-restart for actors will kick-in and meanwhile, any task that it is still healthy, will continue working, so hopefully ready dispatches will still be delivered on time, even if the actor wasn't restarted yet?

If this is the case, maybe you should check tasks health when `_fetch()`ing too, in case some task was aborted, to try to restart it. Right now you are only checking if a task is in the `_scheduled` dict, but not if the tasks are still running or not.

In any case, it is probably worth adding a comment here clarifying how clean up works exactly on different scenarios.

src/frequenz/dispatch/actor.py
```diff
@@ -0,0 +1,279 @@
...
+        ):
+            if task := self._scheduled.pop(dispatch.id, None):
+                task.cancel()
+
+        if dispatch.active and dispatch.id not in self._scheduled:
+            self._scheduled[dispatch.id] = asyncio.create_task(
+                self._schedule_task(dispatch)
+            )
+
+    async def _schedule_task(self, dispatch: Dispatch) -> None:
```
llucax, line 198: My mental picture of how this should have worked was, instead of having one task per dispatch, to just have one task that calculates the next run for all dispatches and task the one that needs to run next, then sleeps only for that time, sends the ready dispatch through the channel, and calculates the next dispatch to be run again.

I think this scheme should be simpler and more memory efficient, as you don't need one task per dispatch. Do you see any advantages with the current approach?

I guess it might be a bit more CPU efficient, as the next dispatch for all dispatches doesn't need to be calculated every time a dispatch is ready, but this could be optimized fairly easily by saving all the next run calculation for all dispatches in a sorted list and only calculate the next run for the dispatch that was just ready. Then we just need to invalidate this list each time dispatches are changed.

src/frequenz/dispatch/actor.py
```diff
@@ -0,0 +1,279 @@
...
+            await asyncio.sleep((next_time - now).total_seconds())
+
+            _logger.info("Dispatch ready: %s", dispatch)
+            await self._ready_dispatch_sender.send(dispatch)
+
+        _logger.info("Dispatch finished: %s", dispatch)
+        self._scheduled.pop(dispatch.id)
+
+    @classmethod
+    def calculate_next_run(cls, dispatch: Dispatch, _from: datetime) -> datetime | None:
```
llucax, line 233: Agreed that is useful to have, I even wonder if we should move this method to the `Dispatch` class instead (so to the client), seems like something anyone using the API client directly probably want to have too, an easy way to know when a dispatch should be run next.

src/frequenz/dispatch/actor.py
```diff
@@ -0,0 +1,279 @@
...
+            byweekday=[
+                _RRULE_WEEKDAY_MAP[weekday]
+                for weekday in dispatch.recurrence.byweekdays
+            ],
+            bymonthday=dispatch.recurrence.bymonthdays,
+            bymonth=dispatch.recurrence.bymonths,
+            interval=dispatch.recurrence.interval,
+        )
+
+        return cast(datetime | None, next_run.after(_from, inc=True))
```
llucax, line 279: Is this `cast` needed because `dateutil` lacks type info or what?

llucax (commented):
OK, quite a few comments, in particular about the design.

> [!IMPORTANT]
> Please let's not discuss the topics here, as discussion could get lengthy. I will create separate issues (or PRs where possible) and link them when I'm done.

pyproject.toml
llucax: * https://github.com/frequenz-floss/frequenz-dispatch-python/issues/13

pyproject.toml
llucax: * https://github.com/frequenz-floss/frequenz-dispatch-python/issues/13

src/frequenz/dispatch/actor.py
llucax: * https://github.com/frequenz-floss/frequenz-dispatch-python/pull/15

llucax: * https://github.com/frequenz-floss/frequenz-dispatch-python/pull/15

src/frequenz/dispatch/actor.py
llucax: * https://github.com/frequenz-floss/frequenz-dispatch-python/pull/15

src/frequenz/dispatch/actor.py
llucax: * https://github.com/frequenz-floss/frequenz-dispatch-python/pull/15

src/frequenz/dispatch/__init__.py
llucax: * https://github.com/frequenz-floss/frequenz-dispatch-python/pull/15

src/frequenz/dispatch/__init__.py
llucax: * https://github.com/frequenz-floss/frequenz-dispatch-python/pull/15

src/frequenz/dispatch/__init__.py
llucax: * https://github.com/frequenz-floss/frequenz-dispatch-python/pull/15

src/frequenz/dispatch/__init__.py
llucax: * https://github.com/frequenz-floss/frequenz-dispatch-python/pull/15

llucax: * https://github.com/frequenz-floss/frequenz-dispatch-python/pull/15

src/frequenz/dispatch/__init__.py
llucax: * https://github.com/frequenz-floss/frequenz-dispatch-python/pull/15

src/frequenz/dispatch/__init__.py
llucax: * https://github.com/frequenz-floss/frequenz-dispatch-python/pull/15

src/frequenz/dispatch/__init__.py
llucax: * https://github.com/frequenz-floss/frequenz-dispatch-python/pull/15

src/frequenz/dispatch/__init__.py
llucax: * https://github.com/frequenz-floss/frequenz-dispatch-python/pull/15

src/frequenz/dispatch/__init__.py
llucax: * https://github.com/frequenz-floss/frequenz-dispatch-python/pull/15

src/frequenz/dispatch/__init__.py
llucax: * https://github.com/frequenz-floss/frequenz-dispatch-python/pull/15

src/frequenz/dispatch/actor.py
llucax: * https://github.com/frequenz-floss/frequenz-dispatch-python/issues/16

src/frequenz/dispatch/actor.py
llucax: * https://github.com/frequenz-floss/frequenz-dispatch-python/issues/17

src/frequenz/dispatch/actor.py
llucax: * https://github.com/frequenz-floss/frequenz-client-dispatch-python/issues/28

src/frequenz/dispatch/__init__.py
llucax: * https://github.com/frequenz-floss/frequenz-dispatch-python/pull/15

src/frequenz/dispatch/actor.py
llucax: This one might be too much work to implement, so I think it is better to discuss it first, and probably it is better to sense the general feeling here first, if it's not a trivial "let's go with this approach" (or let's NOT) before creating an issue, so please go ahead and leave comments here.

src/frequenz/dispatch/actor.py
llucax: This one is too small to create an issue. So also please comment here.

Marenz: Yah, I was getting `src/frequenz/dispatch/actor.py:279: error: Returning Any from function declared to return "datetime | None"  [no-any-return]`

src/frequenz/dispatch/actor.py
Marenz: I think my only concern with your suggestion is that it might accidentally skip dispatches that are very close together, though I can't come up for a good explanation why it would, just an .. gut feeling to pay extra attention on that area when implementing it..

I was also wondering if the awaiting for sending might take a bit of time and thus delays the notification for the next very close dispatch to be executed. What if we waited longer than the next dispatches scheduled time, how do we make sure we don't skip it?

llucax: > I think my only concern with your suggestion is that it might accidentally skip dispatches that are very close together, though I can't come up for a good explanation why it would, just an .. gut feeling to pay extra attention on that area when implementing it..

If done properly, I don't think it should. I didn't invented this, I've seen this (or a similar approach) suggested in [libev](http://software.schmorp.de/pkg/libev.html) as a [very efficient way to implement timers](http://pod.tst.eu/http://cvs.schmorp.de/libev/ev.pod#code_ev_timer_code_relative_and_opti) (see point 4).

> I was also wondering if the awaiting for sending might take a bit of time and thus delays the notification for the next very close dispatch to be executed. What if we waited longer than the next dispatches scheduled time, how do we make sure we don't skip it?

If we await on the send takes a long time, I guess the next notification will be delayed too as it will use the same channel to notify, right? We could do the send in a separate task but then we end up with many tasks again, but I think sends will be serialized in any case because they operate in the same channel, so using a different task is probably not having any real effect.

src/frequenz/dispatch/actor.py
llucax: OK, it is probably best to add a comment when using `cast`, so readers don't have to guess in the future.

PR #15, opened 2024-03-26
Improve project structure

llucax:
- **Move `Dispatcher` to its own module**
- **Expose all public symbols**
- **Remove the `_DispatchEventBase` class**
- **Move dispatch events to their own module**
- **Fix `Dispatcher` example in docstring**
- **Make `Dispatcher` return receiver fetchers**
- **Rename `DispatchActor` to `DispatchingActor`**
- **Use a timer for polling.**
- **Improve `Dispatcher` examples**
- **Rename `updated_dispatches` and `ready_dispatches`**

### DIFF
--- a/src/frequenz/dispatch/_dispatcher.py
+++ b/src/frequenz/dispatch/_dispatcher.py
@@ -3,12 +3,36 @@
 
 """A highlevel interface for the dispatch API."""
 
+import abc
+from typing import Protocol, TypeVar
+
 import grpc.aio
 from frequenz.channels import Broadcast, Receiver
 from frequenz.client.dispatch.types import Dispatch
 
 from frequenz.dispatch._event import DispatchEvent
 from frequenz.dispatch.actor import DispatchActor
+
+ReceivedT = TypeVar("ReceivedT")
+"""The type being received."""
+
+
+class ReceiverFetcher(Protocol[ReceivedT]):
+    """An interface that just exposes a `new_receiver` method."""
+
+    @abc.abstractmethod
+    def new_receiver(
+        self, name: str | None = None, maxsize: int = 50
+    ) -> Receiver[ReceivedT]:
+        """Get a receiver from the channel.
+
+        Args:
+            name: A name to identify the receiver in the logs.
+            maxsize: The maximum size of the receiver.
+
+        Returns:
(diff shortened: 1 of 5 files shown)


llucax:
Up for discussion.

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -60,8 +60,8 @@ async def run():
             service_address = "localhost:50051"
             dispatcher = Dispatcher(microgrid_id, grpc_channel, service_address)
             dispatcher.start()  # this will start the actor
-            dispatch_arrived = dispatcher.updated_dispatches.new_receiver()
-            dispatch_ready = dispatcher.ready_dispatches.new_receiver()
+            events = dispatcher.lifecycle_events.new_receiver()
+            ready = dispatcher.ready_to_execute.new_receiver()
```
Marenz: I was not happy with the names, I like these more :)

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -60,8 +81,19 @@ async def run():
             service_address = "localhost:50051"
             dispatcher = Dispatcher(microgrid_id, grpc_channel, service_address)
             dispatcher.start()  # this will start the actor
-            events = dispatcher.lifecycle_events.new_receiver()
-            ready = dispatcher.ready_to_execute.new_receiver()
+
+            events_receiver = dispatcher.lifecycle_events.new_receiver()
+
+            async for event in events_receiver:
```
Marenz, line 87: This makes me think we should also expose the client as `dispatcher.client` so the user can react to the events in a meaningful way when they want to update or create a new event as a response

Marenz (approved):

Marenz:
Great stuff

src/frequenz/dispatch/_dispatcher.py
llucax: I don't get this. Why do you need the client?

Marenz: Well, anyone who wants to update, delete or create a request needs the client to do so, the high-level API 
doesn't offer those things (and I don't see how it could do that better than the client)

PR #22, opened 2024-04-04
Make `ready_to_execute` channel more general

Marenz:
* Rename `ready_to_execute` channel to `running_status_change` reflecting it's broader purpose.
* Small re-designs and changes about the interfaces in general
* Add own `Dispatch` type with useful functions on top of the original base class

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,38 +7,61 @@
 ## Introduction
 
 A highlevel interface for the dispatch API.
-The interface is made of the dispatch actor which should run once per microgrid.
-It provides two channels for clients:
-- "new_dispatches" for newly created dispatches
-- "ready_dispatches" for dispatches that are ready to be executed
 
-## Example Usage
+See [the documentation](https://frequenz-floss.github.io/frequenz-dispatch-python/v0.1/reference/frequenz/dispatch) for more information.
+
+## Usage
+
+The [`Dispatcher` class](https://frequenz-floss.github.io/frequenz-dispatch-python/v0.1/reference/frequenz/dispatch/#frequenz.dispatch.Dispatcher), the main entry point for the API, provides two channels:
+
+* [Lifecycle events](https://frequenz-floss.github.io/frequenz-dispatch-python/v0.1/reference/frequenz/dispatch/#frequenz.dispatch.Dispatcher.lifecycle_events): A channel that sends a message whenever a [Dispatch][frequenz.dispatch.Dispatch] is created, updated or deleted.
+* [Running status change](https://frequenz-floss.github.io/frequenz-dispatch-python/v0.1/reference/frequenz/dispatch/#frequenz.dispatch.Dispatcher.running_status_change): Sends a dispatch message whenever a dispatch is ready to be executed according to the schedule or the running status of the dispatch changed in a way that could potentially require the actor to start, stop or reconfigure itself.
+
+### Example using the running status change channel
 
 ```python
-    async def run():
-        # dispatch helper sends out dispatches when they are due
-        dispatch_arrived = dispatch_helper.updated_dispatches().new_receiver()
-        dispatch_ready = dispatch_helper.ready_dispatches().new_receiver()
+import os
+import grpc.aio
+from unittest.mock import MagicMock
 
-        async for selected in select(dispatch_ready, dispatch_arrived):
-            if selected_from(selected, dispatch_ready):
-                dispatch = selected.value
-                match dispatch.type:
-                    case DISPATCH_TYPE_BATTERY_CHARGE:
-                        battery_pool = microgrid.battery_pool(dispatch.battery_set, task_id)
-                        battery_pool.set_charge(dispatch.power)
-                ...
-            if selected_from(selected, dispatch_arrived):
-                match selected.value:
-                    case Created(dispatch):
-                        log.info("New dispatch arrived %s", dispatch)
-                        ...
-                    case Updated(dispatch):
-                        log.info("Dispatch updated %s", dispatch)
-                        ...
-                    case Deleted(dispatch):
-                        log.info("Dispatch deleted %s", dispatch)
-                        ...
+async def run():
+    host = os.getenv("DISPATCH_API_HOST", "localhost")
+    port = os.getenv("DISPATCH_API_PORT", "50051")
+
+    service_address = f"{host}:{port}"
+    grpc_channel = grpc.aio.insecure_channel(service_address)
+    microgrid_id = 1
+    dispatcher = Dispatcher(microgrid_id, grpc_channel, service_address)
+    await dispatcher.start()
+
+    actor = MagicMock() # replace with your actor
+
+    changed_running_status_rx = dispatcher.running_status_change.new_receiver()
+
+    async for dispatch in changed_running_status_rx:
+        match dispatch.running("DEMO_TYPE"):
+            case RunningState.RUNNING:
+                print(f"Executing dispatch {dispatch.id}, due on {dispatch.start_time}")
+                if actor.is_running:
+                    actor.reconfigure(
+                        components=dispatch.selector,
+                        run_parameters=dispatch.payload, # custom actor parameters
+                        dry_run=dispatch.dry_run,
+                        until=dispatch.until,
+                    )  # this will reconfigure the actor
+                else:
+                    # this will start a new actor with the given components
+                    # and run it for the duration of the dispatch
+                    actor.start(
+                        components=dispatch.selector,
+                        run_parameters=dispatch.payload, # custom actor parameters
+                        dry_run=dispatch.dry_run,
+                        until=dispatch.until,
+                    )
+            case RunningState.STOPPED:
+                actor.stop()  # this will stop the actor
+            case RunningState.DIFFERENT_TYPE:
+                pass  # dispatch not for this type
 ```
 
 ## Supported Platforms
--- a/src/frequenz/dispatch/_dispatch.py
+++ b/src/frequenz/dispatch/_dispatch.py
@@ -83,11 +83,12 @@ class Dispatch(BaseDispatch):
             running_state_change_synced,
         )
 
-    def set_deleted(self) -> None:
+    def _set_deleted(self) -> None:
         """Mark the dispatch as deleted."""
         object.__setattr__(self, "deleted", True)
 
-    def running_status_notified(self) -> bool:
+    @property
+    def _running_status_notified(self) -> bool:
         """Check that the latest running state change notification was sent.
 
         Returns:
@@ -95,7 +96,7 @@ class Dispatch(BaseDispatch):
         """
         return self.running_state_change_synced == self.update_time
 
-    def set_running_status_notified(self) -> None:
+    def _set_running_status_notified(self) -> None:
         """Mark the latest running state change notification as sent."""
         object.__setattr__(self, "running_state_change_synced", self.update_time)
 
--- a/src/frequenz/dispatch/_dispatch.py
+++ b/src/frequenz/dispatch/_dispatch.py
@@ -1,0 +1,234 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Dispatch type with support for next_run calculation."""
+
+
+import logging
+from dataclasses import dataclass
+from datetime import datetime, timezone
+from typing import Iterator, cast
+
+from dateutil import rrule
+from frequenz.client.dispatch.types import Dispatch as BaseDispatch
+from frequenz.client.dispatch.types import Frequency, Weekday
+
+_logger = logging.getLogger(__name__)
+"""The logger for this module."""
+
+_RRULE_FREQ_MAP = {
+    Frequency.MINUTELY: rrule.MINUTELY,
+    Frequency.HOURLY: rrule.HOURLY,
+    Frequency.DAILY: rrule.DAILY,
+    Frequency.WEEKLY: rrule.WEEKLY,
+    Frequency.MONTHLY: rrule.MONTHLY,
+}
+"""To map from our Frequency enum to the dateutil library enum."""
+
+_RRULE_WEEKDAY_MAP = {
+    Weekday.MONDAY: rrule.MO,
+    Weekday.TUESDAY: rrule.TU,
+    Weekday.WEDNESDAY: rrule.WE,
+    Weekday.THURSDAY: rrule.TH,
+    Weekday.FRIDAY: rrule.FR,
+    Weekday.SATURDAY: rrule.SA,
+    Weekday.SUNDAY: rrule.SU,
+}
+"""To map from our Weekday enum to the dateutil library enum."""
+
+
+@dataclass(frozen=True)
+class Dispatch(BaseDispatch):
+    """Dispatch type with extra functionality."""
+
+    deleted: bool = False
+    """Whether the dispatch is deleted."""
+
+    running_state_change_synced: datetime | None = None
+    """The last time a message was sent about the running state change."""
+
+    def __init__(
+        self,
+        client_dispatch: BaseDispatch,
+        deleted: bool = False,
+        running_state_change_synced: datetime | None = None,
+    ):
+        """Initialize the dispatch.
+
+        Args:
+            client_dispatch: The client dispatch.
+            deleted: Whether the dispatch is deleted.
+            running_state_change_synced: Timestamp of the last running state change message.
+        """
+        super().__init__(**client_dispatch.__dict__)
+        # Work around frozen to set deleted
+        object.__setattr__(self, "deleted", deleted)
+        object.__setattr__(
+            self,
+            "running_state_change_synced",
+            running_state_change_synced,
+        )
+
+    def set_deleted(self) -> None:
+        """Mark the dispatch as deleted."""
+        object.__setattr__(self, "deleted", True)
+
+    def running_status_notified(self) -> bool:
+        """Check that the latest running state change notification was sent.
+
+        Returns:
+            True if the latest running state change notification was sent, False otherwise.
+        """
+        return self.running_state_change_synced == self.update_time
+
+    def set_running_status_notified(self) -> None:
+        """Mark the latest running state change notification as sent."""
+        object.__setattr__(self, "running_state_change_synced", self.update_time)
+
+    def running(self, type_: str) -> bool:
+        """Check if the dispatch is currently supposed to be running.
+
+        Args:
+            type_: The type of the dispatch that should be running.
+
+        Returns:
+            True if the dispatch is currently meant to be running, False otherwise.
+        """
+        if self.type != type_:
+            return False
+
+        if not self.active or self.deleted:
+            return False
+
+        now = datetime.now(tz=timezone.utc)
+        if until := self._until(now):
+            return now < until
+
+        return False
+
+    @property
+    def until(self) -> datetime | None:
+        """Time when the dispatch should end.
+
+        Returns the time that a running dispatch should end.
+        If the dispatch is not running, None is returned.
+
+        Returns:
+            The time when the dispatch should end or None if the dispatch is not running.
+        """
+        if not self.active or self.deleted:
+            return None
+
+        now = datetime.now(tz=timezone.utc)
+        return self._until(now)
+
+    @property
+    # noqa is needed because of a bug in pydoclint that makes it think a `return` without a return
+    # value needs documenting
+    def missed_runs(self) -> Iterator[datetime]:  # noqa: DOC405
+        """Yield all missed runs of a dispatch.
+
+        Yields all missed runs of a dispatch.
+
+        If a running state change notification was not sent in time
+        due to connection issues, this method will yield all missed runs
+        since the last sent notification.
+
+        Returns:
+            A generator that yields all missed runs of a dispatch.
+        """
+        if self.update_time == self.running_state_change_synced:
+            return
+
+        from_time = self.update_time
+        now = datetime.now(tz=timezone.utc)
+
+        while (next_run := self.next_run_after(from_time)) and next_run < now:
+            yield next_run
+            from_time = next_run
+
+    @property
+    def next_run(self) -> datetime | None:
+        """Calculate the next run of a dispatch.
+
+        Returns:
+            The next run of the dispatch or None if the dispatch is finished.
+        """
+        return self.next_run_after(datetime.now(tz=timezone.utc))
+
+    def next_run_after(self, after: datetime) -> datetime | None:
+        """Calculate the next run of a dispatch.
+
+        Args:
+            after: The time to calculate the next run from.
+
+        Returns:
+            The next run of the dispatch or None if the dispatch is finished.
+        """
+        if (
+            not self.recurrence.frequency
+            or self.recurrence.frequency == Frequency.UNSPECIFIED
+        ):
+            if after > self.start_time:
+                return None
+            return self.start_time
+
+        # Make sure no weekday is UNSPECIFIED
+        if Weekday.UNSPECIFIED in self.recurrence.byweekdays:
+            _logger.warning("Dispatch %s has UNSPECIFIED weekday, ignoring...", self.id)
+            return None
+
+        # No type information for rrule, so we need to cast
+        return cast(datetime | None, self._prepare_rrule().after(after, inc=True))
+
+    def _prepare_rrule(self) -> rrule.rrule:
+        """Prepare the rrule object.
+
+        Returns:
+            The rrule object.
+        """
+        count, until = (None, None)
+        if end := self.recurrence.end_criteria:
+            count = end.count
+            until = end.until
+
+        rrule_obj = rrule.rrule(
+            freq=_RRULE_FREQ_MAP[self.recurrence.frequency],
+            dtstart=self.start_time,
+            count=count,
+            until=until,
+            byminute=self.recurrence.byminutes,
+            byhour=self.recurrence.byhours,
+            byweekday=[
+                _RRULE_WEEKDAY_MAP[weekday] for weekday in self.recurrence.byweekdays
+            ],
+            bymonthday=self.recurrence.bymonthdays,
+            bymonth=self.recurrence.bymonths,
+            interval=self.recurrence.interval,
+        )
+
+        return rrule_obj
+
+    def _until(self, now: datetime) -> datetime | None:
+        """Calculate the time when the dispatch should end.
+
+        If no previous run is found, None is returned.
+
+        Args:
+            now: The current time.
+
+        Returns:
+            The time when the dispatch should end or None if the dispatch is not running.
+        """
+        if (
+            not self.recurrence.frequency
+            or self.recurrence.frequency == Frequency.UNSPECIFIED
+        ):
+            return self.start_time + self.duration
+
+        latest_past_start: datetime | None = self._prepare_rrule().before(now, inc=True)
+
+        if not latest_past_start:
+            return None
+
+        return latest_past_start + self.duration
--- a/src/frequenz/dispatch/_dispatch.py
+++ b/src/frequenz/dispatch/_dispatch.py
@@ -7,6 +7,7 @@
 import logging
 from dataclasses import dataclass
 from datetime import datetime, timezone
+from enum import Enum
 from typing import Iterator, cast
 
 from dateutil import rrule
@@ -35,6 +36,19 @@ _RRULE_WEEKDAY_MAP = {
     Weekday.SUNDAY: rrule.SU,
 }
 """To map from our Weekday enum to the dateutil library enum."""
+
+
+class RunningState(Enum):
+    """The running state of a dispatch."""
+
+    RUNNING = "RUNNING"
+    """The dispatch is running."""
+
+    STOPPED = "STOPPED"
+    """The dispatch is stopped."""
+
+    DIFFERENT_TYPE = "DIFFERENT_TYPE"
+    """The dispatch is for a different type."""
 
 
 @dataclass(frozen=True)
@@ -85,26 +99,28 @@ class Dispatch(BaseDispatch):
         """Mark the latest running state change notification as sent."""
         object.__setattr__(self, "running_state_change_synced", self.update_time)
 
-    def running(self, type_: str) -> bool:
+    def running(self, type_: str) -> RunningState:
         """Check if the dispatch is currently supposed to be running.
 
         Args:
             type_: The type of the dispatch that should be running.
 
         Returns:
-            True if the dispatch is currently meant to be running, False otherwise.
+            RUNNING if the dispatch is running,
+            STOPPED if it is stopped,
+            DIFFERENT_TYPE if it is for a different type.
         """
         if self.type != type_:
-            return False
+            return RunningState.DIFFERENT_TYPE
 
         if not self.active or self.deleted:
-            return False
+            return RunningState.STOPPED
 
         now = datetime.now(tz=timezone.utc)
         if until := self._until(now):
-            return now < until
+            return RunningState.RUNNING if now < until else RunningState.STOPPED
 
-        return False
+        return RunningState.STOPPED
 
     @property
     def until(self) -> datetime | None:
--- a/src/frequenz/dispatch/_dispatcher.py
+++ b/src/frequenz/dispatch/_dispatcher.py
@@ -42,59 +42,62 @@ class Dispatcher:
     This class provides a highlevel interface to the dispatch API.
     It provides two channels:
 
-    One that sends a dispatch event message whenever a dispatch is created, updated or deleted.
+    Lifecycle events:
+        A channel that sends a dispatch event message whenever a dispatch
+        is created, updated or deleted.
 
-    The other sends a dispatch message whenever a dispatch is ready to be
-    executed according to the schedule or the running status of the dispatch
-    changed in a way that could potentially require the actor to start, stop or
-    reconfigure itself.
+    Running status change:
+        Sends a dispatch message whenever a dispatch is ready
+        to be executed according to the schedule or the running status of the
+        dispatch changed in a way that could potentially require the consumer to start,
+        stop or reconfigure itself.
 
     Example: Processing running state change dispatches
+        ```python
+        import os
+        import grpc.aio
+        from frequenz.dispatch import Dispatcher, RunningState
+        from unittest.mock import MagicMock
 
-    ```python
-    import os
-    import grpc.aio
-    from unittest.mock import MagicMock
+        async def run():
+            host = os.getenv("DISPATCH_API_HOST", "localhost")
+            port = os.getenv("DISPATCH_API_PORT", "50051")
 
-    async def run():
-        host = os.getenv("DISPATCH_API_HOST", "localhost")
-        port = os.getenv("DISPATCH_API_PORT", "50051")
+            service_address = f"{host}:{port}"
+            grpc_channel = grpc.aio.insecure_channel(service_address)
+            microgrid_id = 1
+            dispatcher = Dispatcher(microgrid_id, grpc_channel, service_address)
+            await dispatcher.start()
 
-        service_address = f"{host}:{port}"
-        grpc_channel = grpc.aio.insecure_channel(service_address)
-        microgrid_id = 1
-        dispatcher = Dispatcher(microgrid_id, grpc_channel, service_address)
-        await dispatcher.start()
+            actor = MagicMock() # replace with your actor
 
-        actor = MagicMock() # replace with your actor
+            changed_running_status = dispatcher.running_status_change.new_receiver()
 
-        changed_running_status_rx = dispatcher.running_status_change.new_receiver()
-
-        async for dispatch in changed_running_status_rx:
-            match dispatch.running("DEMO_TYPE"):
-                case RunningState.RUNNING:
-                    print(f"Executing dispatch {dispatch.id}, due on {dispatch.start_time}")
-                    if actor.is_running:
-                        actor.reconfigure(
-                            components=dispatch.selector,
-                            run_parameters=dispatch.payload, # custom actor parameters
-                            dry_run=dispatch.dry_run,
-                            until=dispatch.until,
-                        )  # this will reconfigure the actor
-                    else:
-                        # this will start a new actor with the given components
-                        # and run it for the duration of the dispatch
-                        actor.start(
-                            components=dispatch.selector,
-                            run_parameters=dispatch.payload, # custom actor parameters
-                            dry_run=dispatch.dry_run,
-                            until=dispatch.until,
-                        )
-                case RunningState.STOPPED:
-                    actor.stop()  # this will stop the actor
-                case RunningState.DIFFERENT_TYPE:
-                    pass  # dispatch not for this type
-    ```
+            async for dispatch in changed_running_status:
+                match dispatch.running("DEMO_TYPE"):
+                    case RunningState.RUNNING:
+                        print(f"Executing dispatch {dispatch.id}, due on {dispatch.start_time}")
+                        if actor.is_running:
+                            actor.reconfigure(
+                                components=dispatch.selector,
+                                run_parameters=dispatch.payload, # custom actor parameters
+                                dry_run=dispatch.dry_run,
+                                until=dispatch.until,
+                            )  # this will reconfigure the actor
+                        else:
+                            # this will start a new actor with the given components
+                            # and run it for the duration of the dispatch
+                            actor.start(
+                                components=dispatch.selector,
+                                run_parameters=dispatch.payload, # custom actor parameters
+                                dry_run=dispatch.dry_run,
+                                until=dispatch.until,
+                            )
+                    case RunningState.STOPPED:
+                        actor.stop()  # this will stop the actor
+                    case RunningState.DIFFERENT_TYPE:
+                        pass  # dispatch not for this type
+        ```
 
     Example: Getting notification about dispatch lifecycle events
         ```python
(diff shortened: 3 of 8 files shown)


src/frequenz/dispatch/_event.py
```diff
@@ -38,3 +38,57 @@ class Deleted:
...
+
+@dataclass(frozen=True)
+class DryRunStop:
+    """A request to stop a running dispatch in dry run mode."""
+
+    dispatch: Dispatch
+    """The dispatch that is being stopped in dry run mode."""
+
+
+DispatchCommandRequest = Start | Stop | DryRunStart | DryRunStop
```
llucax: Mmm, I don't like this approach that much, for a few reasons:

* It seems less resilient: if the dispatch actor crashes, then suddenly all active dispatches will keep running indefinitely. With the current approach we only depend on the dispatch actor to trigger the initial run, but then actors can be more independent.
* Splitting into `dry_run` and normal dispatches will probably end up in code duplication, I like better the approach of just forwarding the `dry_run` flag to the actor doing the actual work so they can simulate what the dispatch should have done as close to the real thing as possible.

llucax: Also semantically, I think the current approach is more precise. You are notified that a dispatch is ready to be executed. It might have been started 5 minutes ago, or 1 hour ago (if it was delayed for some reason, like the LTE connection was down for a while). Start/stop commands sound more *real-time* than things might be.

What happens if a dispatch is created to run now and for one hour, while the LTE connection is down for 1.5h. Then the connection is up, the dispatcher receives the new dispatch, is for 1.5h ago and it should be done now. What do you do, do you ignore it? Do you generate a start and stop command one right after the other? All of those seem to be bad to me. With the current approach you just send a dispatch that is ready to be executed, the consumer can decide if running something that was supposed to run 1.5h ago and finish 0.5h ago makes sense or not.[^1]

I know with the commands you still get the `dispatch` object, so you could do the checks too, but still the interface seems more misleading to me.

[^1]: Actually the issue about if we consider a dispatch that was supposed to be finished by now "ready to be executed" or not affects the current approach too, and it is an open question. Because of what I say above, I think it makes sense to consider it ready to execute, swallowing dispatches silently doesn't seem like a safe approach. I think users will have to always check the start time and duration to know if the dispatch is still active or not and if it should be run even if it should be done by now, or if it should be completely skipped.

Marenz: Well, one motivation behind having `DryRun*` like this was that it is not easy to ignore that way. Otherwise, with relying on the app developer to check the `dry_run` flag is not very safe. It can easily be overlooked or forgotten (I did in my first iteration, thus my motivation to avoid this for future devs).

llucax: I see, that's a fair point. I agree it would be a good thing to make it harder to make mistakes, but I'm not super concerned about missing checking for `dry_run` because it is probably a very useful feature to use during development so I guess it would be hard to go unnoticed for too long. That said, we can find other mechanisms to make it harder to miss the `dry_run` flag, like making the `dry_run` attribute a property and checking if it was ever read, and if it was not read when the `Dispatch` (`ReadyDispatch`?) object is destroyed, we can log a warning or even crash (`assert` if that works in `__del__()` or just `sys.exit()` with an error). We can also add a method `disable_dry_run_check()` in case someone was to purposely ignore it (just reading the property would also be enough, but it might look uglier/more confusing).

Another approach would be to have a separate channel for dry run ready dispatches, but I guess that could lead again to code duplication, or even people now missing the existence of the dry run ready dispatches, which won't prevent the issue.

What do you think?

Marenz: I am honestly most in favor of the approach here, it makes it an non-optional required extra case in the match statement, thus being a very explicit thing the user is now forced to think about right when writing code (as opposed to relying on internal asserts or checks that would throw at runtime as you outlined as some alternatives).

A separate channel would solve it as well yes, but I don't see what we gain with an extra channel that we don't get with a "shared" channel like this PR shows.

Marenz: That being said, I am completely fine with keeping the duration/stop part on the client side and not rely on any Stop being sent.

llucax: We had a chat and decided to mostly keep the original design because doing the split of dry run could be not very easy to use and result in more confusing user code.

We identified an important use case though, and that is notifying users about running dispatch updates, so they don't need to subscribe to the lifecycle channel, which will be very noisy. The idea is to send the dispatch again through the `ready_to_execute` channel each time it is updated. The channel might be renamed to better convey the new semantics.

@Marenz one more thing to have in mind if we are going to add a `dispatch.is_running` property is what to do with dispatches that were supposed to start and end in the past, but there some LTE interruption, and we only know about the existence of that dispatch after it was supposed to be finished. In that case checking for `is_running` won't be very useful, as it should be force, but maybe the dispatch needs to be executed anyway.

I guess this is a broader topic about scheduling. For example, IIRC, cron skips stuff that should have run in the past, but `Anacron` doesn't. [`systemd.timer`](https://www.freedesktop.org/software/systemd/man/latest/systemd.timer.html) must also deal with this, and this is what it does:

> When a system is temporarily put to sleep (i.e. system suspend or hibernation) the realtime clock does not pause. When a calendar timer elapses while the system is sleeping it will not be acted on immediately, but once the system is later resumed it will catch up and process all timers that triggered while the system was sleeping. Note that if a calendar timer elapsed more than once while the system was continuously sleeping the timer will only result in a single service activation. If WakeSystem= (see below) is enabled a calendar time event elapsing while the system is suspended will cause the system to wake up (under the condition the system's hardware supports time-triggered wake-up functionality).

For recurring dispatches there is also the question of what to do if the same dispatch was supposed to run several times, do we run it only once like systemd or do we catch up?

I guess all of this will depend on the type of dispatch, so the dispatch API should probably provide tools to deal with all these options and make documentation clear so users are aware about these possibilities and handle them in the way that makes more sense for each actor.

Marenz: > @Marenz one more thing to have in mind if we are going to add a dispatch.is_running property is what to do with dispatches that were supposed to start and end in the past, but there some LTE interruption, and we only know about the existence of that dispatch after it was supposed to be finished. In that case checking for is_running won't be very useful, as it should be force, but maybe the dispatch needs to be executed anyway.

No, we can't make guarantees like that. The app receives the dispatch when it receives it and evaluates what to do with it. 
If an app(developer) decides they must run all dispatches they have not explicitly previously run (e.g. by logging it in the actor) they can implement it themselves.

For those that care, `update_time` would be the only way to differentiate between a dispatch whose `duration` was reduced vs a dispatch who arrived past it's ending date..

llucax: > No, we can't make guarantees like that. The app receives the dispatch when it receives it and evaluates what to do with it.
> If an app(developer) decides they must run all dispatches they have not explicitly previously run (e.g. by logging it in the actor) they can implement it themselves.
> For those that care, `update_time` would be the only way to differentiate between a dispatch whose `duration` was reduced vs a dispatch who arrived past it's ending date..

I'm not talking about updated dispatches necessarily but also created ones, suppose the LTE connection is down for 4 hours and someone created a recurrent dispatch that should run every hour for 15 minutes just after the LTE connection was down. Now 4 hours later, this library pools once more but this time there is a newly created dispatch. We will send the create event (only one) via the lifecycle channel, that's clear. Now from the user perspective, there is a dispatch that should have been executed 4 times already.

I guess you say that we just wait until the next scheduled dispatch should happen and send only that dispatch via the ready to execute channel? If this is the case, for users that need to ensure all missed scheduled dispatched would have been executed, they will have to do a lot of work.

If this library can track this down and send 4 dispatches (the same object basically) via the ready to execute channel instead, then users can more easily implement this logic, not having to schedule the missed dispatch triggers themselves, but just looking if the dispatch should be running **now** or should have ran in the past and they want to catch up (or not).

Now I'm wondering if the catch-up behavior should be actually a decision made by the actor or the user creating the dispatch, and should be part of the dispatch itself. :thinking:

If it is the later, then this library could act accordingly and send catch-up dispatches through the ready to execute channel or not automatically.

I guess this is a question for Tom, but we probably need to formulate it from scratch, I guess just pinging him here could lead to a lot of confusion if we don't present the problem properly. But first, what do you think?

Marenz: Hmm I suppose with that in mind, it's not a biggy to send late dispatches, the client can stay stupid and just check `.running` or it can be smart and check the past runs etc

Marenz: > 4 dispatches (the same object basically) via the ready to execute channel instead, then users can more easily implement this logic, not having to schedule the missed dispatch triggers themselves

I mean, I would send only one dispatch, how often it should have run can be checked locally then.
We can however provide methods in the dispatch object to make those checks easy.

llucax: Yeah, that sounds good too. Unless we want to make the catch-up behavior part of the `Dispatch` object, in which case I think this library should send several dispatch notification if the catch-up behavior is so.

llucax:
I see this was updated recently. Is it still a thing? Please let me know if you need more input from me.

Marenz:
This is very much a thing, one that will be updated now that the dependencies are merged :)

Marenz:
But apart from rebasing, the code is ready for review

src/frequenz/dispatch/_dispatch.py
```diff
@@ -0,0 +1,226 @@
...
+    def running(self, type_: str) -> bool:
+        """Check if the dispatch is currently supposed to be running.
+
+        Args:
+            type_: The type of the dispatch that should be running.
+
+        Returns:
+            True if the dispatch is currently meant to be running, False otherwise.
+        """
+        if self.type != type_:
```
Marenz, line 114: I am just thinking, maybe this should return `None` for this case, as in "Information not available"

Marenz: Or maybe a three-state enum?

src/frequenz/dispatch/_dispatch.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+import logging
+from dataclasses import dataclass
+from datetime import datetime, timezone
+from typing import Iterator, cast
+
+from dateutil import rrule
+from frequenz.client.dispatch.types import Dispatch as BaseDispatch
+from frequenz.client.dispatch.types import Frequency, Weekday
```
llucax, line 15: (unrelated) I don't remember the rationale about using `.types` here but to me it is still funny-looking that it is not just `from frequenz.client.dispatch import ...`.

llucax: Also unrelated and just purely brainstorming, but it we reversed the namespaces here, it might be more natural to use with google style imports:

```py
from frequenz.dispatch import client

...

class Dispatch(client.Dispatch):
    ...
```

src/frequenz/dispatch/_dispatch.py
```diff
@@ -0,0 +1,226 @@
...
+        running_state_change_synced: datetime | None = None,
+    ):
+        """Initialize the dispatch.
+
+        Args:
+            client_dispatch: The client dispatch.
+            deleted: Whether the dispatch is deleted.
+            running_state_change_synced: Timestamp of the last running state change message.
+        """
+        super().__init__(**client_dispatch.__dict__)
```
llucax, line 77: `dataclasses.asdict()`?

src/frequenz/dispatch/_dispatch.py
```diff
@@ -0,0 +1,226 @@
...
+        object.__setattr__(
+            self,
+            "running_state_change_synced",
+            running_state_change_synced,
+        )
+
+    def set_deleted(self) -> None:
+        """Mark the dispatch as deleted."""
+        object.__setattr__(self, "deleted", True)
+
```
llucax: Is this intended to be used by end users?

src/frequenz/dispatch/_dispatch.py
llucax: What about raising an exception? I think a 3-state enum will be annoying to use and `None` will be, for most cases, the same as using `False`, as people will be still able to use it like `if dispatch.running():` ...

If type changes are not allowed by the API this could only happen due to bugs or intentionally bad players, right?

If you don't like the exception approach I guess returning `None` might be the most convenient option, if it is relevant to anyone, they can do a `if dispatch.running() is None:` and if you don´t care you can use it as a `bool` (the only issue is it will probably lead to people not thinking about the `None` option at all, which could potentially mask issues).

src/frequenz/dispatch/_dispatch.py
```diff
@@ -0,0 +1,226 @@
...
+    def running_status_notified(self) -> bool:
```
llucax: Why not `@property`?

And this also looks like is not intended to be used by the end user. If this is the case, I would make it private.

src/frequenz/dispatch/_dispatch.py
```diff
@@ -0,0 +1,226 @@
...
+        """Check that the latest running state change notification was sent.
+
+        Returns:
+            True if the latest running state change notification was sent, False otherwise.
+        """
+        return self.running_state_change_synced == self.update_time
+
+    def set_running_status_notified(self) -> None:
```
llucax: read-write property?

src/frequenz/dispatch/_dispatch.py
```diff
@@ -0,0 +1,226 @@
...
+        """Mark the latest running state change notification as sent."""
+        object.__setattr__(self, "running_state_change_synced", self.update_time)
+
+    def running(self, type_: str) -> bool:
```
llucax: Why not `@property`?

src/frequenz/dispatch/_dispatch.py
```diff
@@ -0,0 +1,226 @@
...
+            The time when the dispatch should end or None if the dispatch is not running.
+        """
+        if not self.active or self.deleted:
+            return None
+
+        now = datetime.now(tz=timezone.utc)
+        return self._until(now)
+
+    @property
+    def missed_runs(self) -> Iterator[datetime]:  # noqa: DOC405
```
llucax, line 145: Maybe explain te `# noqa`?

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -44,41 +45,71 @@ class Dispatcher:
     One that sends a dispatch event message whenever a dispatch is created, updated or deleted.
 
     The other sends a dispatch message whenever a dispatch is ready to be
-    executed according to the schedule.
+    executed according to the schedule or the running status of the dispatch
+    changed in a way that could potentially require the actor to start, stop or
```
llucax: nitpick: I wouldn't say actor here, the consumer doesn't necessarily have to be an actor, I would say "require the receiver to do some adjustments to the dispatch being ran".

src/frequenz/dispatch/_dispatch.py
llucax: Mmm, looking at the `type_` argument I'm more lost than before, this interface looks very confusing to me. Why are we checking against a type passed by the user at all? Can't the user check this by itself if they are interested in it?

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -44,41 +45,71 @@ class Dispatcher:
...
-                print(f"Executing dispatch {dispatch.id}, due on {dispatch.start_time}")
-                # execute the dispatch
+            await dispatcher.start()
+
+            actor = MagicMock() # replace with your actor
+
+            changed_running_status_rx = dispatcher.running_status_change.new_receiver()
+
+            async for dispatch in changed_running_status_rx:
+                match dispatch.running("DEMO_TYPE"):
```
llucax, line 77: This looks super weird to me...

Why not:

```py
match dispatch.type:
    case "DEMO_TYPE":
        if not dispatch.is_running:
            await actor.stop()
            continue
        print(f"Executing dispatch {dispatch.id}, due on {dispatch.start_time}")
        if actor.is_running:
            ...
        else:
            ...
```

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -95,6 +126,46 @@ async def run():
                     case _ as unhandled:
                         assert_never(unhandled)
         ```
+    Example: Creating a new dispatch and then modifying it. Note that this uses
+    the lower-level `Client` class to create and update the dispatch.
```
llucax: Did you check how this documentation renders? I recommend using a shorter title anyway, and adding a longer description as part of the example:

```suggestion
    Example: Creating a new dispatch and then modifying it
    
        Note that this uses the lower-level `Client` class to create and update the dispatch.
```

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -95,6 +126,46 @@ async def run():
...
+            service_address = f"{host}:{port}"
+            grpc_channel = grpc.aio.insecure_channel(service_address)
+            microgrid_id = 1
+            dispatcher = Dispatcher(microgrid_id, grpc_channel, service_address)
+            await dispatcher.start()  # this will start the actor
+
+            # Create a new dispatch
+            new_dispatch = await dispatcher.client.create(
+                microgrid_id=microgrid_id,
+                _type="ECHO_FREQUENCY",  # replace with your own type
```
llucax, line 159: BTW, why `_type`? `type` is not a reserved word, `flake8`/`pylint` triggers a warning if it is used by I think we should silence it and use it anyway in this case.

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -107,34 +178,69 @@ def __init__(
...
         Returns:
             A new receiver for new dispatches.
         """
-        return self._updated_channel
+        return self._lifecycle_events_channel
 
     @property
-    def ready_to_execute(self) -> ReceiverFetcher[Dispatch]:
-        """Return ready dispatches receiver.
+    def running_status_change(self) -> ReceiverFetcher[Dispatch]:
```
llucax, line 218: I'm not happy about this name, since we are *streamming* `Dispatch` objects, I think it should be named accordingly, like which types of dispatch objects are put here. I think currently running dispatches are put here, or *active* dispatches.

So what about either just `active`, or `currently_active`? Either that, or we call it something like `_events` like `lifecycle_events` and wrap the dispatches inside some other object that tells which kind of update is it.

For me it looks simpler to just use `active`, but it might be actually misleading, as it is actually a sort of stream of status changes as you say, but then we should have some wrapper object and track the changes and report the changes instead of just sending the dispatch object itself. I'm really unsure about this though, as it is a lot of work and the actual current needs are pretty well covered by this interface, so maybe it makes more sense to go with the simple `active` approach and clarify in the docs, it seems like no name will be clear and good enough to convey all that users will need to do to properly handle dispatches that are changing while running.

src/frequenz/dispatch/actor.py
```diff
@@ -114,18 +89,28 @@ async def _fetch(self) -> None:
...
             _logger.info("Fetching dispatches for microgrid %s", self._microgrid_id)
-            async for dispatch in self._client.list(microgrid_id=self._microgrid_id):
-                self._dispatches[dispatch.id] = dispatch
+            async for client_dispatch in self._client.list(
+                microgrid_id=self._microgrid_id
+            ):
+                dispatch = Dispatch(client_dispatch)
 
+                self._dispatches[dispatch.id] = Dispatch(client_dispatch)
                 old_dispatch = old_dispatches.pop(dispatch.id, None)
```
llucax, line 98: Be careful with this, if the actor crashes, you might lose this dispatch in the list of `old_dispatches`, not sure if it would be a problem or not.

src/frequenz/dispatch/actor.py
```diff
@@ -114,18 +89,28 @@ async def _fetch(self) -> None:
...
                 if not old_dispatch:
                     self._update_dispatch_schedule(dispatch, None)
                     _logger.info("New dispatch: %s", dispatch)
-                    await self._updated_dispatch_sender.send(Created(dispatch=dispatch))
+                    await self._lifecycle_updates_sender.send(
+                        Created(dispatch=dispatch)
+                    )
```
llucax, line 104: What if it is just created but already ready to run? Shouldn't it be sent to the running state change channel too, or will that be handled by a separate task taking care of the scheduling?

llucax (commented):
I like this in general, but I have a few minor comments and some more fundamental, like about the `running()` function and the channel name.

src/frequenz/dispatch/_dispatch.py
Marenz: Every actor receives every dispatch, so they have to check if it is for them.

To make sure they _have_ to check, the `running()` method requires the type and if the types don't match no action at all is expected. 

If the users don't handle the three-state enum, they instead need to filter the dispatch. Seems at the end the same amount of work to me, only that this way forces the user and makes it impossible to forget

src/frequenz/dispatch/_dispatch.py
Marenz: because it has a parameter

src/frequenz/dispatch/_dispatcher.py
Marenz: Active seems wrong as well, given that it also sends deleted and specifically inactive dispatches, too...

src/frequenz/dispatch/actor.py
Marenz: Yes, this happens in `_schedule_task()` called from two lines higher `_update_dispatch_schedule()`

src/frequenz/dispatch/_dispatcher.py
Marenz: > like which types of dispatch objects are put here. I think currently running dispatches are put here, or active dispatches.

Any "type" that makes a receiver start, stop or reconfigure.

> So what about either just active, or currently_active?

It includes deleted, inactive and finished dispatches, as well. 

> or we call it something like _events like lifecycle_events 

We could call it `running_status_events` :shrug: 

> and wrap the dispatches inside some other object that tells which kind of update is it.

Not sure what help that is, there is no specific "kind" or "type" of update.
Well, or if you want, you could categorize it as

1) request to start
2) request to stop
3) request to use different configuration

all those cases are covered by the simple usage example already shown, basically for 1) and 3) you check if you're running, if yes, reconfigure, if no, start. (or stop if it's 2)

src/frequenz/dispatch/_dispatch.py
llucax: Probably for a separate PR, but I was thinking that maybe this is the wrong approach. If we know any users will only be interested in a sub-set of dispatch types, then it would probably be better to have an interface like:

```py
class FcrDispatch(Dispatch):
    def __init__(self, dispatch: Dispatch) -> None:
        ...
        # Probably parsing the payload here too, using pydantic or some other schema
        # we could even offer tools for this
        super().__init__(dispatch, FcrPayloadSchema())

class OtherDispatch(Dispatch):
    ...
        
async for dispatch in dispatcher.active.new_receiver({"FCR": FcrDispatch, "OTHER": OtherDispatch}):
    match dispatch:
        case FcrDispatch() as fcr_dispatch:
            ...
        case OtherDispatch() as other_dispatch:
            ...
        case _ as unhandled:
            assert_never(unhandled)
```

This way you are forced to check the type, and you can have a more specific interface for the types you care about.

So the `new_receiver()` method would become more complex, it needs to do the filtering and the conversion of the general `Dispatch` type to the specific one provided by the user, and be generic. It might be a challenge to do it properly. I guess worse case we can always only just do it with a simple `T` and then use: `dispatcher.active.new_receiver[FcrDispatch | OtherDispatch]({"FCR": FcrDispatch, "OTHER": OtherDispatch})` which is a bit too verbose, but well.

src/frequenz/dispatch/_dispatch.py
llucax: Yeah, sorry, I wrote that before realizing it was used also to check the type.

src/frequenz/dispatch/_dispatcher.py
llucax: > It includes deleted, inactive and finished dispatches, as well.

Yeah, but considering both the previous and new state, then the dispatch was active (either it needs to be active now, or it was active when the change was received). It is a broad definition of active but I think it could still work given that it is very complex what we are trying to describe and no sensible symbol name will be able to fully describe it.

> We could call it `running_status_events` 🤷

Yeah, sure, but in that case I don't think we should stream `Dispatch` objects, we should stream some `RunningStatusEvent`, which makes the interface more complicated, because what do you put there?

> Not sure what help that is, there is no specific "kind" or "type" of update. Well, or if you want, you could categorize it as
> 
>     1. request to start
>     2. request to stop
>     3. request to use different configuration

Mmmm, maybe. For some reason I'm not convinced to add so much meaning to an update. What we are really receiving is an update to the dispatch, which could end up in those 3 points or not. But I don't know exactly why I don't like this mapping, maybe is just a general feeling of leaking abstractions, maybe you are right and this approach is OK.

But then, when the dispatch finishes normally you won't get a notification this way, right? Or will you? I thought we only send something here if the duration changed, or it was disabled, or something else that makes the dispatch not be active anymore, but users should still make sure the dispatch is stopped after the duration even if they don't receive any notification, right? My fear here is that if users believe they will always be notified about a stopping event, then they will not look into the duration anymore.

Actually thinking about it, the actor could send a stop request and keep the timer(s) itself to trigger the event after the duration. I guess the only risk here is the dispatch actor crashing and the users not being notified about the stop. We could potentially handle this correctly and recover the timers after the crash, so only an infinite restart loop could a real problem, but maybe in that case is reasonable to assume everything went to hell.

Another issue with handling the duration timer in the dispatch actor is that we'll have to set a timer for every dispatch, even if nobody cares about it. Let's say for example a FCR dispatch has an invalid payload parameter, so we don't run it, we'll still have a timer in the dispatch actor for it. If we combine it with the feature of automatically filtering by type in the `Dispatcher`, then I guess this problem is not that bad, as it should be rare that someone requested for a type of Dispatch and then they don't really run it.

OK, if we go this route I think it could work: We have `running_status_events`, which streams something like `Start | Stop | Update` events, we provide type filtering via the `new_receiver()` and we make the actor handle the dispatch timer. This will make the dispatch actor much more complex, but I guess it might be worth it if we remove that complexity from the user...

Marenz: > But then, when the dispatch finishes normally you won't get a notification this way, right?

correct.

> Or will you? I thought we only send something here if the duration changed, or it was disabled, or something else that makes the dispatch not be active anymore, but users should still make sure the dispatch is stopped after the duration even if they don't receive any notification, right? My fear here is that if users believe they will always be notified about a stopping event, then they will not look into the duration anymore.

Though I feel we could offer this as a client side `ActorDispatcher` class like [here](https://github.com/frequenz-io/frequenz-actor-fcr/pull/349/files#diff-17c28766ef849b0c33eace241ef4a4d28aa7e399b96d1e569d3bf8bd9cc1f2b2R22) so that the end-user would really just have the `start()` and `stop()` and then `reconfigure()` or so functions to care about.

> Actually thinking about it, the [..]

Well, or we go that way, of course. I do like the `new_receiver(filter)` idea, that makes whatever I've been doing with `running()` obsolete

src/frequenz/dispatch/_dispatch.py
Marenz: I mean, it's the same, dataclasses specifically set the dict attribute

src/frequenz/dispatch/_dispatch.py
Marenz: No, will make private

src/frequenz/dispatch/_dispatch.py
llucax: It is less hacky and I wouldn't be surprised if there is more magic going on. I don't see any reason to use the hacky approach if the official way is enough :)

src/frequenz/dispatch/_dispatcher.py
llucax: * https://github.com/frequenz-floss/frequenz-client-dispatch-python/issues/45

README.md
```diff
@@ -7,38 +7,98 @@
...
-It provides two channels for clients:
-- "new_dispatches" for newly created dispatches
-- "ready_dispatches" for dispatches that are ready to be executed
 
-## Example Usage
+A small overview of the most important classes in this package:
+
+* [Dispatcher][frequenz.dispatch.Dispatcher]: The entry point for the API.
+* [Dispatch][frequenz.dispatch.Dispatch]: A dispatch type with lots of useful extra functionality.
+* [Created][frequenz.dispatch.Created], [Updated][frequenz.dispatch.Updated], [Deleted][frequenz.dispatch.Deleted]: Dispatch event types.
```
llucax: You shouldn't use cross-reference syntax in the README, it doesn't render well in GitHub, which will be the primary place where this is seen.

image

I recommend to only provide an example in the README and then pointing to the generated documentation for more details. This kind of overview should be in the **User Guide** IMHO (i.e. probably the `__init__.py` docstring).

You can see the [channels repo](https://github.com/frequenz-floss/frequenz-channels-python/tree/v1.x.x/docs/user-guide) as an example to how to structure a user guide where most of the docs still lives in the docstrings, so it is still accessible in code editors.

README.md
```diff
@@ -7,38 +7,98 @@
...
-                        log.info("Dispatch updated %s", dispatch)
-                        ...
-                    case Deleted(dispatch):
-                        log.info("Dispatch deleted %s", dispatch)
-                        ...
+import os
+import grpc.aio
+from unittest.mock import MagicMock
+
+async def run():
```
llucax, line 27: We also have a couple of very long examples in the channels readme, and in retrospect I think it was a bad idea, it is not helpful to have a very quick feel of the library, I suggest including only this example in the readme and leave the lifecycle example for the user guide.

README.md
```diff
@@ -7,38 +7,98 @@
...
+
+    service_address = f"{host}:{port}"
+    grpc_channel = grpc.aio.insecure_channel(service_address)
+    microgrid_id = 1
+    dispatcher = Dispatcher(microgrid_id, grpc_channel, service_address)
+    await dispatcher.start()
+
+    actor = MagicMock() # replace with your actor
+
+    changed_running_status_rx = dispatcher.running_status_change.new_receiver()
```
llucax, line 39: I wouldn't use the `_rx` suffix in the example, IMHO it is a bit cryptic, I would drop it or use `_receiver` like in the lifecycle example.

```suggestion
    changed_running_status = dispatcher.running_status_change.new_receiver()
```

src/frequenz/dispatch/_dispatch.py
llucax: Waiting for feedback on this one, I really think we should use the publicly exposed functions, in python one should usually not use symbols starting with `__` unless it is really necessary. `__dict__` is not even documented in https://docs.python.org/3/library/dataclasses.html.

src/frequenz/dispatch/_dispatch.py
llucax: So we decided to go with this approach, and also the class approach on top of it, so simple dispatch executing actors should be trivial to implement, and if you need something a bit more complex, there is still a pretty high-level interface that takes most of the complexity out.

But we'll do this in a separate PR. @Marenz do you want to create a new issue with this info or do you plan to start implementing it immediately after this is merged?

src/frequenz/dispatch/_dispatch.py
llucax: Maybe? :pray: :laughing:

src/frequenz/dispatch/_dispatch.py
```diff
@@ -0,0 +1,243 @@
...
+    @property
+    def _running_status_notified(self) -> bool:
+        """Check that the latest running state change notification was sent.
+
+        Returns:
+            True if the latest running state change notification was sent, False otherwise.
+        """
+        return self.running_state_change_synced == self.update_time
+
+    def _set_running_status_notified(self) -> None:
```
llucax, line 99:
```suggestion
    @_running_status_notified.setter
    def _running_status_notified(self) -> None:
```

src/frequenz/dispatch/_dispatcher.py
llucax: Also pending, it should be a quick fix.

llucax:
@Marenz I updated the title to make it more suitable as a commit message. Can you please also update the PR description to be suitable as a commit message body? This will end up in the merge commit.

src/frequenz/dispatch/_dispatch.py
Marenz: I'll implement it right a way one FCR dispatching is up and running

src/frequenz/dispatch/_dispatch.py
Marenz: I am not sure what to say.... the whole thing is your fault anyway, wanting to use `Iterator` :p

> src/frequenz/dispatch/_dispatch.py:144:1: DOC405 Method `Dispatch.missed_runs` has both "return" and "yield" statements. Please use Generator[YieldType, SendType, ReturnType] as the return ty
pe annotation, and put your yield type in YieldType and return type in ReturnType. More details in https://jsh9.github.io/pydoclint/notes_generator_vs_iterator.html

src/frequenz/dispatch/_dispatch.py
Marenz: I tried using it and it actually breaks the tests because it converts sub-members also to dicts...

```
Pdb) client_dispatch.__dict__
{'id': 1, 'microgrid_id': 26, 'type': '99346', 'start_time': datetime.datetime(2024, 5, 17, 10, 45, 58, tzinfo=datetime.timezone.utc), 'duration': datetime.timedelta(days=5, seconds=9001), 'selector': [66, 63, 52, 39, 62], 'active': True, 'dry_run': True, 'payload': {'key_3': 'c', 'key_0': 75.58042041572239, 'key_1': True, 'key_2': False}, 'recurrence': RecurrenceRule(frequency=<Frequency.WEEKLY: 4>, interval=67, end_criteria=None, byminutes=[45, 52, 50, 42, 40, 0], byhours=[15, 10, 7, 23, 10, 22, 2, 6, 18], byweekdays=[<Weekday.TUESDAY: 2>, <Weekday.SUNDAY: 7>, <Weekday.TUESDAY: 2>], bymonthdays=[15, 3, 3, 11, 29, 17, 30, 16], bymonths=[5]), 'create_time': datetime.datetime(2024, 5, 6, 15, 27, 50, 359784, tzinfo=datetime.timezone.utc), 'update_time': datetime.datetime(2024, 5, 6, 15, 27, 50, 359784, tzinfo=datetime.timezone.utc)}
(Pdb) dataclasses.asdict(client_dispatch)
{'id': 1, 'microgrid_id': 26, 'type': '99346', 'start_time': datetime.datetime(2024, 5, 17, 10, 45, 58, tzinfo=datetime.timezone.utc), 'duration': datetime.timedelta(days=5, seconds=9001), 'selector': [66, 63, 52, 39, 62], 'active': True, 'dry_run': True, 'payload': {'key_3': 'c', 'key_0': 75.58042041572239, 'key_1': True, 'key_2': False}, 'recurrence': {'frequency': <Frequency.WEEKLY: 4>, 'interval': 67, 'end_criteria': None, 'byminutes': [45, 52, 50, 42, 40, 0], 'byhours': [15, 10, 7, 23, 10, 22, 2, 6, 18], 'byweekdays': [<Weekday.TUESDAY: 2>, <Weekday.SUNDAY: 7>, <Weekday.TUESDAY: 2>], 'bymonthdays': [15, 3, 3, 11, 29, 17, 30, 16], 'bymonths': [5]}, 'create_time': datetime.datetime(2024, 5, 6, 15, 27, 50, 359784, tzinfo=datetime.timezone.utc), 'update_time': datetime.datetime(2024, 5, 6, 15, 27, 50, 359784, tzinfo=datetime.timezone.utc)}
```

clearly `asdict()` is doing too much magic :P

src/frequenz/dispatch/_dispatch.py
Marenz: I am not sure how I'd call this setter given that I provide no value...

README.md
Marenz: Left only the basic example in now

README.md
Marenz: fixed links

src/frequenz/dispatch/actor.py
Marenz: Hmm it might not be a problem to just re-run all dispatches that are currently supposed to be running?

src/frequenz/dispatch/_dispatch.py
llucax: Mmm, OK, I'm still not happy about using an undocumented method starting with `__` but if replacing it is not trivial I'm OK with leaving it *temporarily* (i.e. until all blows up because `__dict__` is removed :laughing: ).

src/frequenz/dispatch/_dispatch.py
llucax: OK, sorry, I missed there was no argument. I think it would be better to either take an argument or maybe rename to something like `_mark_as_sent` or something conveying what this is actually doing better. But it is a minor thing if we plan to change this in the future, I'm fine with merging as is if it is going to be changed soon.

src/frequenz/dispatch/_dispatch.py
llucax: Mmm, OK, I'm still not happy about using an undocumented method starting with `__` but if replacing it is not trivial I'm OK with leaving it *temporarily* (i.e. until all blows up because `__dict__` is removed :laughing: ).

src/frequenz/dispatch/_dispatch.py
```diff
@@ -0,0 +1,245 @@
...
+        """
+        if not self.active or self.deleted:
+            return None
+
+        now = datetime.now(tz=timezone.utc)
+        return self._until(now)
+
+    @property
+    # noqa is to silence warning about Return & Yield statements in the same
+    # function
```
llucax: I would say something like:
```suggestion
    # noqa is needed because of a bug in pydoclint that makes it think a `return` with a return
    # value needs documenting
```

src/frequenz/dispatch/_dispatch.py
llucax: See the new comment. We should probably also report the issue, it shouldn't be hard to fix I guess, just treating a `return` without a value specially.

llucax (dismissed):
I'd be happy if the `noqa` comment is updated, but all are minor comments, so approving in case you want to merge as-is.

Marenz:
@llucax anything missing for a merge? :)

llucax:
Oh, sorry, I didn't see it was updated, will have a look now.

README.md
```diff
@@ -7,38 +7,61 @@
...
 
-## Example Usage
+See [the documentation](https://frequenz-floss.github.io/frequenz-dispatch-python/v0.1/reference/frequenz/dispatch) for more information.
+
+## Usage
+
+The [`Dispatcher` class](https://frequenz-floss.github.io/frequenz-dispatch-python/v0.1/reference/frequenz/dispatch/#frequenz.dispatch.Dispatcher), the main entry point for the API, provides two channels:
+
+* [Lifecycle events](https://frequenz-floss.github.io/frequenz-dispatch-python/v0.1/reference/frequenz/dispatch/#frequenz.dispatch.Dispatcher.lifecycle_events): A channel that sends a message whenever a [Dispatch][frequenz.dispatch.Dispatch] is created, updated or deleted.
+* [Running status change](https://frequenz-floss.github.io/frequenz-dispatch-python/v0.1/reference/frequenz/dispatch/#frequenz.dispatch.Dispatcher.running_status_change): Sends a dispatch message whenever a dispatch is ready to be executed according to the schedule or the running status of the dispatch changed in a way that could potentially require the actor to start, stop or reconfigure itself.
```
llucax, line 18: About these links, we were using redirect for version aliases because GitHub didn't support symlinks. With redirects we can't use direct links to an alias, but with symlinks we can. You could update the `alias_type` [configuration](https://github.com/jimporter/mike?tab=readme-ov-file#configuration) from `redirect` to `symlink` and then link to the `latest` alias for example, instead of `v0.1`.

That said, any linking to any particular version will be bad as things are now, because the README is also rendered as the home page of the project, and if you are visiting the docs for any particular version, the links in that version home page will point to a different version.

This is why I'm more inclined to **not** provide *deep links* to docs in the README, just explain in plain text and add a link to the docs home page (without any version specification, just https://frequenz-floss.github.io/frequenz-dispatch-python/) for people looking for more info.

That, or stop including the README entirely in the [`docs/index.md`](https://github.com/frequenz-floss/frequenz-dispatch-python/blob/v0.x.x/docs/index.md) and do some partial inclusions (and leave stuff with deep links to docs out of the included parts) like we do with channels:

* https://raw.githubusercontent.com/frequenz-floss/frequenz-channels-python/v1.x.x/README.md
* https://raw.githubusercontent.com/frequenz-floss/frequenz-channels-python/v1.x.x/docs/index.md

Happy to leave this for a future PR though.

src/frequenz/dispatch/_dispatch.py
```diff
@@ -0,0 +1,251 @@
...
+            The time when the dispatch should end or None if the dispatch is not running.
+        """
+        if not self.active or self.deleted:
+            return None
+
+        now = datetime.now(tz=timezone.utc)
+        return self._until(now)
+
+    @property
+    # noqa is needed because of a bug in pydoclint that makes it think a `return` with a return
```
llucax:
```suggestion
    # noqa is needed because of a bug in pydoclint that makes it think a `return` without a return
```

:grimacing:

llucax (commented):
The typo in the comment is super minor but very confusing if left as is, so ideally fix that and we are set.

README.md
Marenz: Let's do that in the next iteration that we already plan :) I am afraid to add more stuff here currently

src/frequenz/dispatch/_dispatch.py
Marenz: fixed

llucax (approved):

PR #23, opened 2024-04-24
Update frequenz dependencies

Marenz:


src/frequenz/dispatch/_dispatcher.py
```diff
@@ -13,22 +13,22 @@
 from frequenz.dispatch._event import DispatchEvent
 from frequenz.dispatch.actor import DispatchingActor
 
-ReceivedT = TypeVar("ReceivedT")
```
Marenz, line 16: The channels dependency changed to covariant types, so we have to follow suit

tests/test_frequenz_dispatch.py
```diff
@@ -207,7 +198,7 @@ async def test_existing_dispatch_deleted(
         case Created(dispatch) | Updated(dispatch):
             assert False, "Expected a deleted event"
         case Deleted(dispatch):
-            sample.create_time = dispatch.create_time
```
Marenz, line 210: Create time is now set through line 191 `sample = ...` which returns the created dispatch.

tests/test_frequenz_dispatch.py
```diff
@@ -63,32 +62,16 @@ class ActorTestEnv:
...
     microgrid_id = randint(1, 100)
 
     actor = DispatchingActor(
         microgrid_id=microgrid_id,
         grpc_channel=MagicMock(),
         svc_addr="localhost",
-        updated_dispatch_sender=YieldingSender(updated_dispatches),
-        ready_dispatch_sender=YieldingSender(ready_dispatches),
+        updated_dispatch_sender=updated_dispatches.new_sender(),
+        ready_dispatch_sender=ready_dispatches.new_sender(),
```
shsms, line 74: you don't need the sleep anymore?

shsms (commented):
Just one comment/question.  Other than that, are the use of `replace` instead of setting values, and the removal of the `YieldingSender` related to the sdk update?

Either way, it might be nicer to put them in separate commits, so you can explain the reasons for them a bit more.

Marenz:
> the use of replace instead of setting values

This is because the client-dispatch update changed `Dispatch` classes to `frozen`

> and the removal of the YieldingSender

That is a good question. As the tests show, it doesn't seem to be needed anymore, I am not sure about the exact reason though. Maybe it was never actually required to begin with and I added it to solve a problem but actually a different changed fixed it?

shsms:
split into 3 commits, why don't you want more stats in: https://github.com/frequenz-floss/frequenz-dispatch-python/graphs/contributors

Marenz:
> and the removal of the YieldingSender

It was removed because `send()` used to be async but isn't anymore in the latest version

tests/test_frequenz_dispatch.py
Marenz: I sleep very little nowadays :laughing:

shsms:
`send()` in channels repo is still async, not sure if you mean something else, bit LGTM otherwise.

Marenz:
> send() in channels repo is still async, not sure if you mean something else, bit LGTM otherwise.

I don't even know at this point, but it works now :shrug:

pyproject.toml
llucax, line 1: The commit "Update pyproject dependencies" has some conflict comments in the commit message.

src/frequenz/dispatch/_dispatcher.py
llucax: Yeah, by now maybe we should move `ReceiverFetcher` to `channels`, it proven to be useful.

pyproject.toml
```diff
@@ -90,6 +88,7 @@ dev-pytest = [
   "pytest-asyncio == 0.23.6",
   "async-solipsism == 0.6",
   "time-machine == 2.14.1",
+  "sybil == 6.0.3",
```
llucax, line 92: Maybe add a comment about why `sybil` is pinned now?

llucax (commented):
"Adjust code to new channels version" commit message needs an update, because as Sahas pointed out, `send()` didn't change, so the reason why `YieldingSender` is not needed anymore should be another one. If we don't know exactly I guess it is better to just say that instead.

llucax (approved):

PR #26, opened 2024-05-02
Remove explicit dependency on sybil

llucax:
This dependency is used by repo-config, not us directly.


llucax:
I will force-merge this so we can pull in dependabot updates.

PR #34, opened 2024-06-20
Add key authentication

Marenz:
Update examples to use key and TLS

### DIFF
--- a/src/frequenz/dispatch/_dispatcher.py
+++ b/src/frequenz/dispatch/_dispatcher.py
@@ -175,7 +175,12 @@ class Dispatcher:
     """
 
     def __init__(
-        self, microgrid_id: int, grpc_channel: grpc.aio.Channel, svc_addr: str
+        self,
+        *,
+        microgrid_id: int,
+        grpc_channel: grpc.aio.Channel,
+        svc_addr: str,
+        key: str,
     ):
         """Initialize the dispatcher.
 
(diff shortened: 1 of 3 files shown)


Marenz:
Based on https://github.com/frequenz-floss/frequenz-client-dispatch-python/pull/56

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -175,20 +175,26 @@ async def run():
...
             grpc_channel: The gRPC channel.
             svc_addr: The service address.
+            key: The key to access the service.
         """
         self._running_state_channel = Broadcast[Dispatch](name="running_state_change")
         self._lifecycle_events_channel = Broadcast[DispatchEvent](
             name="lifecycle_events"
         )
-        self._client = Client(grpc_channel, svc_addr)
+        self._client = Client(grpc_channel=grpc_channel, svc_addr=svc_addr, key=key)
```
ela-kotulska-frequenz, line 223: Looks good to me.
Just side note that 3 out of 4 arguments to Dispatcher constructor are really arguments to Client constructor. 
And they are just passed to Client. And every description for these arguments is repeated in Client. 
Maybe we could have single struct like ClientArgs in the future?

ela-kotulska-frequenz (approved):

src/frequenz/dispatch/_dispatcher.py
Marenz: That could work, though we want to change to the [base-client](https://github.com/frequenz-floss/frequenz-client-base-python/blob/v0.x.x/src/frequenz/client/base/client.py#L116) where it will do this job for us

PR #41, opened 2024-07-23
Fix running state behavior with delete

Marenz:
When a dispatch is deleted, the running state change notification telling the actor that it is no longer running will only be sent if there are no other dispatches of the same type running.

Signed-off-by: Mathias L. Baumann <mathias.baumann@frequenz.com>

### DIFF
--- a/src/frequenz/dispatch/actor.py
+++ b/src/frequenz/dispatch/actor.py
@@ -34,7 +34,7 @@ _logger = logging.getLogger(__name__)
 """The logger for this module."""
 
 
-class DispatchingActor(Actor):
+class DispatchingActor(Actor):  # pylint: disable=too-many-instance-attributes
     """Dispatch actor.
 
     This actor is responsible for handling dispatches for a microgrid.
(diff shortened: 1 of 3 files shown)


src/frequenz/dispatch/actor.py
```diff
@@ -65,6 +65,7 @@ def __init__(
 
         self._client = client
         self._dispatches: dict[int, Dispatch] = {}
+        self._dispatch_type_ids_map: dict[str, list[int]] = {}
```
llucax: I would add documentation about what this is for, as it is not trivial to figure it out.

llucax (commented):
So the previous behaviour was that a stop event was sent even when there were other dispatches running for this type, so the dispatch would have stopped when it shouldn't, right?

Marenz:
> So the previous behaviour was that a stop event was sent even when there were other dispatches running for this type, so the dispatch would have stopped when it shouldn't, right?

correct

llucax (commented):
I think we should not apply this, as if we need to be able to handle overlapping dispatches that have different parameters, we don't get notified when one of these dispatches should stop.

PR #42, opened 2024-07-26
Bump the cross-ref object DB version to match the dependency version

llucax:
The `frequenz-client-dispatch-python` dependency was updated to `0.5`, but the cross-ref object DB version was not updated to match it, so cross linking is sending users to the wrong version.


Marenz (approved):
This will totally never be forgotten again

llucax:
We need a ci check for this

PR #51, opened 2024-09-04
Update to latest SDK and dispatch-client versions

Marenz:


daniel-zullo-frequenz (approved):

pyproject.toml
```diff
@@ -39,9 +39,9 @@ dependencies = [
   # Make sure to update the version for cross-referencing also in the
   # mkdocs.yml file when changing the version here (look for the config key
   # plugins.mkdocstrings.handlers.python.import)
-  "frequenz-sdk == 1.0.0-rc801",
+  "frequenz-sdk == 1.0.0-rc900",
```
llucax, line 42: Probably not necessary, as the idea is not to make patch releases normally, but you could also use `frequenz-sdk >= 1.0.0-rc900, < 1.0.0-rc1000`

llucax (approved):

pyproject.toml
Marenz: I wasn't sure, because it was a pre-release and the tools have been... special in their treatment of those. So I'd preferred giving the exact version

PR #52, opened 2024-09-05
Fix running state behavior with delete

Marenz:
When a dispatch is deleted, the running state change notification telling the actor that it is no longer running will only be sent if there are no other dispatches of the same type running.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,15 +6,8 @@
 
 ## Upgrading
 
-- The dispatch high level interface now depends on `frequenz-sdk` version `v1.0.0-rc900`.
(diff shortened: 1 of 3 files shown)


Marenz:
@llucax I assume this PR would break jacks battery charging actor?

llucax:
Isn't this a deja vu of https://github.com/frequenz-floss/frequenz-dispatch-python/pull/41?

llucax:
Again, I think the most flexible approach would be to track dispatches by dispatch ID, not by type, and then have helper functions/classes to do any sort of merging we could need, like merging dispatches with the same type to be considered one instance.

llucax:
Not sure where we talked about this, but we talked about it. I think in some PR.

Marenz:
Yeah, I was just seeing the branch and couldn't remember :D

PR #53, opened 2024-09-11
Send stop signal when duration is reached

Marenz:
- **Reset release notes**
- **Send STOPPED message when duration is reached**

### DIFF
--- a/src/frequenz/dispatch/actor.py
+++ b/src/frequenz/dispatch/actor.py
@@ -183,10 +183,20 @@ class DispatchingActor(Actor):
             _logger.info("Dispatch %s scheduled for %s", dispatch.id, next_time)
             await asyncio.sleep((next_time - now).total_seconds())
 
-            _logger.info("Dispatch ready: %s", dispatch)
+            _logger.info("Dispatch %s executing...", dispatch)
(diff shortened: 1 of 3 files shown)


src/frequenz/dispatch/actor.py
```diff
@@ -183,10 +183,20 @@ def next_run_info() -> tuple[datetime, datetime] | None:
...
+            _logger.info("Dispatch %s executing...", dispatch)
             await self._running_state_change_sender.send(dispatch)
 
-        _logger.info("Dispatch finished: %s", dispatch)
+            # Wait for the duration of the dispatch if set
+            if dispatch.duration:
+                _logger.info(
+                    "Dispatch %s running for %s", dispatch.id, dispatch.duration
+                )
+                await asyncio.sleep(dispatch.duration.total_seconds())
```
llucax, line 194: Are you sure a sleep will work? Doesn't this block receiving/running any other dispatch while the current dispatch is running? Shouldn't you use timers instead or this code runs on a separate task for every dispatch that is currently running?

Marenz: Yes ,currently we have one task for each dispatch running.

llucax (approved):

PR #54, opened 2024-09-11
Dispatch Managing Actor

Marenz:
- **Reset release notes**
- **Add dispatch runner**

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -10,7 +10,7 @@
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+* We now provide the `DispatchManagingActor` class, a class to manage actors based on incoming dispatches.
 
 ## Bug Fixes
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -165,6 +165,7 @@ disable = [
 [tool.pytest.ini_options]
 testpaths = ["tests", "src"]
 asyncio_mode = "auto"
+asyncio_default_fixture_loop_scope = "function"
 required_plugins = ["pytest-asyncio", "pytest-mock"]
 
 [tool.mypy]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,9 +40,9 @@ dependencies = [
   # mkdocs.yml file when changing the version here (look for the config key
   # plugins.mkdocstrings.handlers.python.import)
   "frequenz-sdk >= 1.0.0-rc900, < 1.0.0-rc1000",
-  "frequenz-channels >= 1.1.0, < 2.0.0",
+  "frequenz-channels >= 1.2.0, < 2.0.0",
 #  "frequenz-client-dispatch >= 0.6.0, < 0.7.0",
-  "frequenz-client-dispatch @ git+https://github.com/frequenz-floss/frequenz-client-dispatch-python.git@v0.x.x",
+  "frequenz-client-dispatch @ git+https://github.com/frequenz-floss/frequenz-client-dispatch-python.git@refs/pull/87/head",
 ]
 dynamic = ["version"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,10 +39,9 @@ dependencies = [
   # Make sure to update the version for cross-referencing also in the
   # mkdocs.yml file when changing the version here (look for the config key
   # plugins.mkdocstrings.handlers.python.import)
-  "frequenz-sdk >= 1.0.0-rc900, < 1.0.0-rc1000",
+  "frequenz-sdk == 1.0.0-rc900, < 1.0.0-rc1000",
   "frequenz-channels >= 1.2.0, < 2.0.0",
-#  "frequenz-client-dispatch >= 0.6.0, < 0.7.0",
-  "frequenz-client-dispatch @ git+https://github.com/frequenz-floss/frequenz-client-dispatch-python.git@refs/pull/87/head",
+  "frequenz-client-dispatch >= 0.7.0, < 0.8.0",
 ]
 dynamic = ["version"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,8 @@ dependencies = [
   # plugins.mkdocstrings.handlers.python.import)
   "frequenz-sdk >= 1.0.0-rc900, < 1.0.0-rc1000",
   "frequenz-channels >= 1.1.0, < 2.0.0",
-  "frequenz-client-dispatch >= 0.6.0, < 0.7.0",
+#  "frequenz-client-dispatch >= 0.6.0, < 0.7.0",
+  "frequenz-client-dispatch @ git+https://github.com/frequenz-floss/frequenz-client-dispatch-python.git@v0.x.x",
 ]
 dynamic = ["version"]
 
--- a/src/frequenz/dispatch/_dispatch.py
+++ b/src/frequenz/dispatch/_dispatch.py
@@ -119,11 +119,11 @@ class Dispatch(BaseDispatch):
 
         now = datetime.now(tz=timezone.utc)
 
+        if now < self.start_time:
+            return RunningState.STOPPED
         # A dispatch without duration is always running once it started
         if self.duration is None:
-            if self.start_time <= now:
-                return RunningState.RUNNING
-            return RunningState.STOPPED
+            return RunningState.RUNNING
 
         if until := self._until(now):
             return RunningState.RUNNING if now < until else RunningState.STOPPED
--- a/src/frequenz/dispatch/_dispatch.py
+++ b/src/frequenz/dispatch/_dispatch.py
@@ -117,6 +117,10 @@ class Dispatch(BaseDispatch):
         if not self.active or self.deleted:
             return RunningState.STOPPED
 
+        # A dispatch without duration is always running
+        if self.duration is None:
+            return RunningState.RUNNING
+
         now = datetime.now(tz=timezone.utc)
         if until := self._until(now):
             return RunningState.RUNNING if now < until else RunningState.STOPPED
--- a/src/frequenz/dispatch/_dispatch.py
+++ b/src/frequenz/dispatch/_dispatch.py
@@ -117,11 +117,14 @@ class Dispatch(BaseDispatch):
         if not self.active or self.deleted:
             return RunningState.STOPPED
 
-        # A dispatch without duration is always running
-        if self.duration is None:
-            return RunningState.RUNNING
-
         now = datetime.now(tz=timezone.utc)
+
+        # A dispatch without duration is always running once it started
+        if self.duration is None:
+            if self.start_time <= now:
+                return RunningState.RUNNING
+            return RunningState.STOPPED
+
         if until := self._until(now):
             return RunningState.RUNNING if now < until else RunningState.STOPPED
 
@@ -189,6 +192,7 @@ class Dispatch(BaseDispatch):
         if (
             not self.recurrence.frequency
             or self.recurrence.frequency == Frequency.UNSPECIFIED
+            or self.duration is None  # Infinite duration
         ):
             if after > self.start_time:
                 return None
@@ -240,7 +244,13 @@ class Dispatch(BaseDispatch):
 
         Returns:
             The time when the dispatch should end or None if the dispatch is not running.
+
+        Raises:
+            ValueError: If the dispatch has no duration.
         """
+        if self.duration is None:
+            raise ValueError("_until: Dispatch has no duration")
+
         if (
             not self.recurrence.frequency
             or self.recurrence.frequency == Frequency.UNSPECIFIED
--- a/src/frequenz/dispatch/_managing_actor.py
+++ b/src/frequenz/dispatch/_managing_actor.py
@@ -5,7 +5,7 @@
 
 import logging
 from dataclasses import dataclass
-from typing import Any
+from typing import Any, Set
 
 from frequenz.channels import Receiver, Sender
 from frequenz.client.dispatch.types import ComponentSelector
@@ -93,20 +93,20 @@ class DispatchManagingActor(Actor):
 
         status_receiver = dispatcher.running_status_change.new_receiver()
 
-        dispatch_runner = DispatchManagingActor(
+        managing_actor = DispatchManagingActor(
             actor=my_actor,
             dispatch_type="EXAMPLE",
             running_status_receiver=status_receiver,
             updates_sender=dispatch_updates_channel.new_sender(),
         )
 
-        await asyncio.gather(dispatcher.start(), dispatch_runner.start())
+        await asyncio.gather(dispatcher.start(), managing_actor.start())
     ```
     """
 
     def __init__(
         self,
-        actor: Actor,
+        actor: Actor | Set[Actor],
         dispatch_type: str,
         running_status_receiver: Receiver[Dispatch],
         updates_sender: Sender[DispatchUpdate] | None = None,
@@ -114,34 +114,36 @@ class DispatchManagingActor(Actor):
         """Initialize the dispatch handler.
 
         Args:
-            actor: The actor to manage.
+            actor: A set of actors or a single actor to manage.
             dispatch_type: The type of dispatches to handle.
             running_status_receiver: The receiver for dispatch running status changes.
             updates_sender: The sender for dispatch events
         """
         super().__init__()
         self._dispatch_rx = running_status_receiver
-        self._actor = actor
+        self._actors = frozenset([actor] if isinstance(actor, Actor) else actor)
         self._dispatch_type = dispatch_type
         self._updates_sender = updates_sender
 
-    def _start_actor(self) -> None:
-        """Start the actor."""
-        if self._actor.is_running:
-            _logger.warning("Actor %s is already running", self._actor.name)
-        else:
-            self._actor.start()
+    def _start_actors(self) -> None:
+        """Start all actors."""
+        for actor in self._actors:
+            if actor.is_running:
+                _logger.warning("Actor %s is already running", actor.name)
+            else:
+                actor.start()
 
-    async def _stop_actor(self, msg: str) -> None:
-        """Stop the actor.
+    async def _stop_actors(self, msg: str) -> None:
+        """Stop all actors.
 
         Args:
-            msg: The message to be passed to the actor being stopped.
+            msg: The message to be passed to the actors being stopped.
         """
-        if self._actor.is_running:
-            await self._actor.stop(msg)
-        else:
-            _logger.warning("Actor %s is not running", self._actor.name)
+        for actor in self._actors:
+            if actor.is_running:
+                await actor.stop(msg)
+            else:
+                _logger.warning("Actor %s is not running", actor.name)
 
     async def _run(self) -> None:
         """Wait for dispatches and handle them."""
@@ -158,7 +160,7 @@ class DispatchManagingActor(Actor):
         match running:
             case RunningState.STOPPED:
                 _logger.info("Stopped by dispatch %s", dispatch.id)
-                await self._stop_actor("Dispatch stopped")
+                await self._stop_actors("Dispatch stopped")
             case RunningState.RUNNING:
                 if self._updates_sender is not None:
                     _logger.info("Updated by dispatch %s", dispatch.id)
@@ -171,7 +173,7 @@ class DispatchManagingActor(Actor):
                     )
 
                 _logger.info("Started by dispatch %s", dispatch.id)
-                self._start_actor()
+                self._start_actors()
             case RunningState.DIFFERENT_TYPE:
                 _logger.debug(
                     "Unknown dispatch! Ignoring dispatch of type %s", dispatch.type
--- a/src/frequenz/dispatch/_managing_actor.py
+++ b/src/frequenz/dispatch/_managing_actor.py
@@ -1,0 +1,178 @@
+# License: All rights reserved
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Helper class to manage actors based on dispatches."""
+
+import logging
+from dataclasses import dataclass
+from typing import Any
+
+from frequenz.channels import Receiver, Sender
+from frequenz.client.dispatch.types import ComponentSelector
+from frequenz.sdk.actor import Actor
+
+from ._dispatch import Dispatch, RunningState
+
+_logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True, kw_only=True)
+class DispatchUpdate:
+    """Event emitted when the dispatch changes."""
+
+    components: ComponentSelector
+    """Components to be used."""
+
+    dry_run: bool
+    """Whether this is a dry run."""
+
+    options: dict[str, Any]
+    """Additional options."""
+
+
+class DispatchManagingActor(Actor):
+    """Helper class to manage actors based on dispatches.
+
+    Example usage:
+
+    ```python
+    import os
+    import asyncio
+    from frequenz.dispatch import Dispatcher, DispatchManagingActor, DispatchUpdate
+    from frequenz.client.dispatch.types import ComponentSelector
+    from frequenz.client.common.microgrid.components import ComponentCategory
+
+    from frequenz.channels import Receiver, Broadcast
+
+    class MyActor(Actor):
+        def __init__(self, updates_channel: Receiver[DispatchUpdate]):
+            super().__init__()
+            self._updates_channel = updates_channel
+            self._dry_run: bool
+            self._options : dict[str, Any]
+
+        async def _run(self) -> None:
+            while True:
+                update = await self._updates_channel.receive()
+                print("Received update:", update)
+
+                self.set_components(update.components)
+                self._dry_run = update.dry_run
+                self._options = update.options
+
+        def set_components(self, components: ComponentSelector) -> None:
+            match components:
+                case [int(), *_] as component_ids:
+                    print("Dispatch: Setting components to %s", components)
+                case [ComponentCategory.BATTERY, *_]:
+                    print("Dispatch: Using all battery components")
+                case unsupported:
+                    print(
+                        "Dispatch: Requested an unsupported selector %r, "
+                        "but only component IDs or category BATTERY are supported.",
+                        unsupported,
+                    )
+
+    async def run():
+        url = os.getenv("DISPATCH_API_URL", "grpc://fz-0004.frequenz.io:50051")
+        key  = os.getenv("DISPATCH_API_KEY", "some-key")
+
+        microgrid_id = 1
+
+        dispatcher = Dispatcher(
+            microgrid_id=microgrid_id,
+            server_url=url,
+            key=key
+        )
+
+        # Create update channel to receive dispatch update events pre-start and mid-run
+        dispatch_updates_channel = Broadcast[DispatchUpdate](name="dispatch_updates_channel")
+
+        # Start actor and give it an dispatch updates channel receiver
+        my_actor = MyActor(dispatch_updates_channel.new_receiver())
+
+        status_receiver = dispatcher.running_status_change.new_receiver()
+
+        dispatch_runner = DispatchManagingActor(
+            actor=my_actor,
+            dispatch_type="EXAMPLE",
+            running_status_receiver=status_receiver,
+            updates_sender=dispatch_updates_channel.new_sender(),
+        )
+
+        await asyncio.gather(dispatcher.start(), dispatch_runner.start())
+    ```
+    """
+
+    def __init__(
+        self,
+        actor: Actor,
+        dispatch_type: str,
+        running_status_receiver: Receiver[Dispatch],
+        updates_sender: Sender[DispatchUpdate] | None = None,
+    ) -> None:
+        """Initialize the dispatch handler.
+
+        Args:
+            actor: The actor to manage.
+            dispatch_type: The type of dispatches to handle.
+            running_status_receiver: The receiver for dispatch running status changes.
+            updates_sender: The sender for dispatch events
+        """
+        super().__init__()
+        self._dispatch_rx = running_status_receiver
+        self._actor = actor
+        self._dispatch_type = dispatch_type
+        self._updates_sender = updates_sender
+
+    def _start_actor(self) -> None:
+        """Start the actor."""
+        if self._actor.is_running:
+            _logger.warning("Actor %s is already running", self._actor.name)
+        else:
+            self._actor.start()
+
+    async def _stop_actor(self, msg: str) -> None:
+        """Stop the actor.
+
+        Args:
+            msg: The message to be passed to the actor being stopped.
+        """
+        if self._actor.is_running:
+            await self._actor.stop(msg)
+        else:
+            _logger.warning("Actor %s is not running", self._actor.name)
+
+    async def _run(self) -> None:
+        """Wait for dispatches and handle them."""
+        async for dispatch in self._dispatch_rx:
+            await self._handle_dispatch(dispatch=dispatch)
+
+    async def _handle_dispatch(self, dispatch: Dispatch) -> None:
+        """Handle a dispatch.
+
+        Args:
+            dispatch: The dispatch to handle.
+        """
+        running = dispatch.running(self._dispatch_type)
+        match running:
+            case RunningState.STOPPED:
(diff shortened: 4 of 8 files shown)


tests/test_runner.py
```diff
@@ -0,0 +1,171 @@
...
+        runner_actor=runner_actor,
+        running_status_sender=channel.new_sender(),
+        configuration_receiver=config_channel.new_receiver(),
+    )
+
+    await runner_actor.stop()
+
+
+# Disable for _handle_dispatch access
+# pylint: disable=protected-access
```
daniel-zullo-frequenz: These lines might no longer be needed

tests/test_runner.py
```diff
@@ -0,0 +1,171 @@
...
+import time_machine
+from frequenz.channels import Broadcast, Receiver, Sender
+from frequenz.client.dispatch.test.generator import DispatchGenerator
+from frequenz.client.dispatch.types import Frequency
+from frequenz.sdk.actor import Actor
+from pytest import fixture
+
+from frequenz.dispatch import Dispatch, DispatchConfigurationEvent, DispatchRunnerActor
+
+# pylint: disable=protected-access
```
daniel-zullo-frequenz: Probably not needed

src/frequenz/dispatch/_actor_runner.py
```diff
@@ -0,0 +1,187 @@
...
+            await self._handle_dispatch(dispatch=dispatch)
+
+    async def _handle_dispatch(self, dispatch: Dispatch) -> None:
+        """Handle a dispatch.
+
+        Args:
+            dispatch: The dispatch to handle.
+
+        Returns:
+            The running state.
```
daniel-zullo-frequenz: The function returns None :thinking:

daniel-zullo-frequenz (commented):
Just a few really minor comments to check for, LGTM otherwise

src/frequenz/dispatch/_actor_runner.py
```diff
@@ -0,0 +1,187 @@
...
+    """Event emitted when the dispatch configuration changes."""
+
+    components: ComponentSelector
+    """Components to be used."""
+
+    dry_run: bool
+    """Whether this is a dry run."""
+
+    payload: dict[str, Any]
+    """Additional payload."""
```
llucax: I'm still not 100% sold on this, I still feel that it might be better to send an actual `Dispatch` object, possibly a merged one (i.e. if we are merging all dispatches with different IDs but the same type, for example, and we have id=1 start=10:00 end=11:00 and id=2 start=11:00 end=None, that would be merged into a dispatch with start=10:00 and end=None).

But since that would be harder to implement I'm OK with starting with this simple approach and see if we need something more complicated in the future.

src/frequenz/dispatch/_actor_runner.py
```diff
@@ -0,0 +1,187 @@
...
+                self._dry_run = config.dry_run
+                self._payload = config.payload
+
+        def set_components(self, components: ComponentSelector) -> None:
+            match components:
+                case [int(), *_] as component_ids:
+                    print("Dispatch: Setting components to %s", components)
+                case [ComponentCategory.BATTERY, *_]:
+                    print("Dispatch: Using all battery components")
+                case _ as unsupported:
```
llucax: I used this scheme until I realized that is exactly the same as `case unsupported:`.

src/frequenz/dispatch/_actor_runner.py
```diff
@@ -0,0 +1,187 @@
...
+        microgrid_id = 1
+
+        dispatcher = Dispatcher(
+            microgrid_id=microgrid_id,
+            server_url=url,
+            key=key
+        )
+
+        # Create config channel to receive (re-)configuration events pre-start and mid-run
+        config_channel = Broadcast[DispatchConfigurationEvent](name="config_channel")
```
llucax: I know this is a separate repo, but I was confused for a second with the config manager. I would try to make a more explicit separation, maybe call this `dispatch_reconfiguration_channel` (or `dispatch_updates_channel`). I actually think `update` is a better name here, this is not about configuration, but about receiving updates to a running dispatch.

src/frequenz/dispatch/_actor_runner.py
```diff
@@ -0,0 +1,187 @@
...
+
+        # Start actor and supporting actor, give each a config channel receiver
+        my_actor = MyActor(config_channel.new_receiver())
+        supporting_actor = MagicMock(config_channel.new_receiver())
```
llucax: I don't get what this `supporting_actor` is.

src/frequenz/dispatch/_actor_runner.py
```diff
@@ -0,0 +1,187 @@
...
+
+        status_receiver = dispatcher.running_status_change.new_receiver()
+
+        dispatch_handler = DispatchRunnerActor(
```
llucax: I would call this `dispatch_runner = DispatchRunnerActor(...)` for consistency.

src/frequenz/dispatch/_actor_runner.py
```diff
@@ -0,0 +1,187 @@
...
+            configuration_sender=config_channel.new_sender(),
+        )
+
+        await asyncio.gather(dispatcher.start(), dispatch_handler.start())
+    ```
+    """
+
+    def __init__(
+        self,
+        actors: frozenset[Actor],
```
llucax: Do we have usecases for passing multiple actors to this actor?

src/frequenz/dispatch/_actor_runner.py
```diff
@@ -0,0 +1,187 @@
...
+                await actor.stop(msg)
+            else:
+                _logger.warning("Actor %s is not running", actor.name)
+
+    async def _run(self) -> None:
+        """Wait for dispatches and handle them."""
+        while True:
+            _logger.info("Waiting for dispatch...")
+            dispatch = await self._dispatch_rx.receive()
+            await self._handle_dispatch(dispatch=dispatch)
```
llucax: I would use the more idiomatic:

```py
        async for dispatch in self._dispatch_rx:
            await self._handle_dispatch(dispatch=dispatch)
```

And then add a log when a dispatch update is received (like "dispatch update received, processing...", probably with debug level) and finished processing (like "dispatch update processed"). You lose the explicit "waiting" info message, but it is also implicit that it is waiting once it finished processing a dispatch. You also get an automatic info log when the actor is started (via the `Actor` class).

src/frequenz/dispatch/_actor_runner.py
```diff
@@ -0,0 +1,187 @@
...
+            dispatch: The dispatch to handle.
+
+        Returns:
+            The running state.
+        """
+        running = dispatch.running(self._dispatch_type)
+        match running:
+            case RunningState.STOPPED:
+                _logger.info("Stopping dispatch...")
+                await self._stop_actors("Dispatch stopped")
```
llucax: Maybe "Stopped by a dispatch".

src/frequenz/dispatch/_actor_runner.py
```diff
@@ -0,0 +1,187 @@
...
+from frequenz.client.dispatch.types import ComponentSelector
+from frequenz.sdk.actor import Actor
+
+from ._dispatch import Dispatch, RunningState
+
+_logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True, kw_only=True)
+class DispatchConfigurationEvent:
```
llucax: I would call this `DispatchUpdate`. I think calling this configuration is a bit confusing because we use *configuration* for something else, and also I think even should be "created", "updated", etc. For me this class only holds a *dispatch update*.

llucax (commented):
I would say all comments are mostly bike shedding, but in particular I think not using *configuration* and use *dispatch update* is important, I think it can get a bit confusing for people dealing with dispatches and configuration (as in config manager) too. I really think we should replace all `*configuration* mentions to *update*/*dispatch update*.

Also I found the example a bit confusing, which ideally it shouldn't be the case, the example should help make things less confusing :P

src/frequenz/dispatch/_actor_runner.py
Marenz: Dunno, @daniel-zullo-frequenz seems to have required it, that part is from him :D

Marenz:
I don't like `update` much in this context either though. How about `DispatchParameterUpdate`

src/frequenz/dispatch/_actor_runner.py
daniel-zullo-frequenz: Yes, there is a case where we start/stop more than 2 actors. Is there any reason why you'd like to handle only a single actor?

llucax: It feels like it is more natural to map one actor to handling one dispatch, but I'm not strongly opposed. I would like to hear about the details on your use case thought, maybe in private to avoid spamming the PR too much?

llucax:
> I don't like `update` much in this context either though. How about `DispatchParameterUpdate`

Why don't you like `update`? Adding `Parameter` doesn't make it better IMHO, there is no concept of a dispatch parameter in the API. Another option would be `change` instead of `update`, but for me `update` sounds more accurate in this context, this is why I wonder what you don't like about `update`.

RELEASE_NOTES.md
```diff
@@ -10,7 +10,7 @@
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+* We now provide the `DispatchRunnerActor` class, a class to manage actors based on incoming dispatches.
```
llucax: Given this description, `DispatchManagingActor` doesn't sound like a bad option either. But again, bike shedding, I think it is only marginally better.

llucax (commented):
I see there was not much progress or replies on my comments. I guess you are focusing on getting shit to work, which sounds totally reasonable, but just saying that I didn't review much in this round.

src/frequenz/dispatch/_actor_runner.py
```diff
@@ -0,0 +1,184 @@
...
+    """Components to be used."""
+
+    dry_run: bool
+    """Whether this is a dry run."""
+
+    payload: dict[str, Any]
+    """Additional payload."""
+
+
+class DispatchRunnerActor(Actor):
```
llucax: bikeshedding: I'm more and more convinced this is more a `DispatchManagingActor`.

src/frequenz/dispatch/_actor_runner.py
```diff
@@ -0,0 +1,184 @@
...
+class DispatchUpdate:
+    """Event emitted when the dispatch configuration changes."""
+
+    components: ComponentSelector
+    """Components to be used."""
+
+    dry_run: bool
+    """Whether this is a dry run."""
+
+    payload: dict[str, Any]
```
llucax: bikeshedding: Payload sounds weird to me. For a whole dispatch it makes sense (or at least it was the best we could find) but in this context maybe something like `options` is more intuitive? To be honest I think `options` might be a good choice for the specs too.

src/frequenz/dispatch/_actor_runner.py
```diff
@@ -0,0 +1,184 @@
...
+from frequenz.sdk.actor import Actor
+
+from ._dispatch import Dispatch, RunningState
+
+_logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True, kw_only=True)
+class DispatchUpdate:
+    """Event emitted when the dispatch configuration changes."""
```
llucax: Following my quest to remove "config" from dispatch...

```suggestion
    """Event emitted when the dispatch changes."""
```

src/frequenz/dispatch/_actor_runner.py
```diff
@@ -0,0 +1,184 @@
...
+            self._payload: dict[str, Any]
+
+        async def _run(self) -> None:
+            while True:
+                config = await self._config_channel.receive()
+                print("Received config:", config)
+
+                self.set_components(config.components)
+                self._dry_run = config.dry_run
+                self._payload = config.payload
```
llucax:
```suggestion
        def __init__(self, updates_channel: Receiver[DispatchUpdate]):
            super().__init__()
            self._updates_channel = updates_channel
            self._dry_run: bool
            self._payload: dict[str, Any]

        async def _run(self) -> None:
            while True:
                update = await self._updates_channel.receive()
                print("Received update:", update)

                self.set_components(update.components)
                self._dry_run = update.dry_run
                self._payload = update.payload
```

src/frequenz/dispatch/_actor_runner.py
```diff
@@ -0,0 +1,184 @@
...
+    class MyActor(Actor):
+        def __init__(self, config_channel: Receiver[DispatchUpdate]):
+            super().__init__()
+            self._config_channel = config_channel
+            self._dry_run: bool
+            self._payload: dict[str, Any]
+
+        async def _run(self) -> None:
+            while True:
+                config = await self._config_channel.receive()
```
llucax: And again:

```py
        async for dispatch in self._dispatch_rx:
            await self._handle_dispatch(dispatch=dispatch)
```

But actually this example doesn't look very realistic, I guess the actor should do something else than just *listening* to updates. Maybe it is worth adding some other logic to represent that, maybe implement a `ClockActor` that ticks every X time, where X is a dispatch option? So we have a `select()` instead with a `timer` (printing the time) and the `updates_channel`.

src/frequenz/dispatch/_actor_runner.py
```diff
@@ -0,0 +1,184 @@
...
+
+        microgrid_id = 1
+
+        dispatcher = Dispatcher(
+            microgrid_id=microgrid_id,
+            server_url=url,
+            key=key
+        )
+
+        # Create config channel to receive (re-)configuration events pre-start and mid-run
```
llucax:
```suggestion
        # Create updates channel to receive update events pre-start and mid-run
```

src/frequenz/dispatch/_actor_runner.py
```diff
@@ -0,0 +1,184 @@
...
+        dispatch_updates_channel = Broadcast[DispatchUpdate](name="dispatch_updates_channel")
+
+        # Start actor and supporting actor, give each a config channel receiver
```
llucax:
```suggestion
        # Start actor and supporting actor, give each a updates channel receiver
```

src/frequenz/dispatch/_actor_runner.py
```diff
@@ -0,0 +1,184 @@
...
+        # Optional: One (or more) additional actors can be added to the set
+        supporting_actor = MagicMock(dispatch_updates_channel.new_receiver())
+
+        status_receiver = dispatcher.running_status_change.new_receiver()
+
+        dispatch_runner = DispatchRunnerActor(
+            actors=frozenset([my_actor, supporting_actor]),
+            dispatch_type="EXAMPLE",
+            running_status_receiver=status_receiver,
+            configuration_sender=dispatch_updates_channel.new_sender(),
```
llucax:
```suggestion
            updates_sender=dispatch_updates_channel.new_sender(),
```

src/frequenz/dispatch/_actor_runner.py
```diff
@@ -0,0 +1,184 @@
...
+        await asyncio.gather(dispatcher.start(), dispatch_runner.start())
+    ```
+    """
+
+    def __init__(
+        self,
+        actors: frozenset[Actor],
+        dispatch_type: str,
+        running_status_receiver: Receiver[Dispatch],
+        configuration_sender: Sender[DispatchUpdate] | None = None,
```
llucax:
```suggestion
        updates_sender: Sender[DispatchUpdate] | None = None,
```

src/frequenz/dispatch/_actor_runner.py
```diff
@@ -0,0 +1,184 @@
...
+    ) -> None:
+        """Initialize the dispatch handler.
+
+        Args:
+            actors: The actors to handle.
+            dispatch_type: The type of dispatches to handle.
+            running_status_receiver: The receiver for dispatch running status changes.
+            configuration_sender: The sender for dispatch configuration events
```
llucax:
```suggestion
            updates_sender: The sender for dispatch updates.
```

src/frequenz/dispatch/_actor_runner.py
```diff
@@ -0,0 +1,184 @@
...
+        """
+        super().__init__()
+        self._dispatch_rx = running_status_receiver
+        self._actors = actors
+        self._dispatch_type = dispatch_type
+        self._configuration_sender = configuration_sender
```
llucax:
```suggestion
        self._updates_sender = updates_sender
```

src/frequenz/dispatch/_actor_runner.py
```diff
@@ -0,0 +1,184 @@
...
+            dispatch: The dispatch to handle.
+        """
+        running = dispatch.running(self._dispatch_type)
+        match running:
+            case RunningState.STOPPED:
+                _logger.info("Stopped by dispatch %s", dispatch.id)
+                await self._stop_actors("Dispatch stopped")
+            case RunningState.RUNNING:
+                if self._configuration_sender is not None:
+                    _logger.info("Updated configuration by dispatch %s", dispatch.id)
```
llucax:
```suggestion
                if self._updates_sender is not None:
                    _logger.info("Updated by dispatch %s", dispatch.id)
```

src/frequenz/dispatch/_dispatch.py
```diff
@@ -185,6 +192,7 @@ def next_run_after(self, after: datetime) -> datetime | None:
         if (
             not self.recurrence.frequency
             or self.recurrence.frequency == Frequency.UNSPECIFIED
+            or self.duration is None  # Infinite duration
```
llucax, line 195: For a different PR (or to convert to an issue in the client repo):

At the python client level we could actually very easily use special values to represent `INFINITE` and `INSTANTANEOUS`, so the code gets more readable (without the need for comments).

We could set the type of `duration: timedelta | Duration` where duration is a enum for example, then we can also use `match` like this:

```py
match dispatch.duration:
    case Duration.INFINITE: ...
    case Duration.INSTANTANEOUS: ...
    case timedelta(): ... (here we know it is > 0)
    case unexpected: assert_never(unexpected)
```

Taking it even a bit further, we could have a `PositiveTimedelta` object that even guarantee the time is positive.

src/frequenz/dispatch/actor.py
```diff
@@ -59,31 +45,97 @@ def __init__(
...
+                        )
+                        self._dispatches[dispatch.id] = dispatch
+                        await self._lifecycle_updates_sender.send(
+                            Updated(dispatch=dispatch)
+                        )
+                    case Event.DELETED:
+                        self._dispatches.pop(dispatch.id)
+                        await self._update_dispatch_schedule_and_notify(None, dispatch)
+
+                        dispatch._set_deleted()  # pylint: disable=protected-access
```
llucax, line 112: Also for a separate issue: if this function is used externally, then it should be public. If we want to restrict access only to certain users, one trick I've been using is to use either a `Protocol` for the public interface (where you don't expose the `set_deleted()` method) or just plain inheritance (`Dispatch` without `set_delete()` and a `DeletableDispatch` that inherits from it; internally you use `DeletableDispatch` but when returning it to the users, you return a plain `Dispatch`).

This actually made me think why are are storing state in a immutable object? That looks weird.
Another approach, since the `Dispatch`'s `deleted` `bool` and `running_state_change_synced` are only attached as mutable properties for internal **state** tracking, it might make more sense to separate the **state** from the immutable dispatch, and have a `DispatchState` **mutable** class and a separate `dict` to map dispatch ID to its state.

You can pass the state object to the `Dispatch` constructor and store it as a private property and expose *read-only* properties to access the state, if it needs to be accessed externally, and store the `dict` and mutate it from this actor only.

tests/test_frequenz_dispatch.py
```diff
@@ -15,18 +15,30 @@
...
+)
 from frequenz.dispatch.actor import DispatchingActor
 
 
-# This method replaces the event loop for all tests in the file.
-@fixture
-def event_loop_policy() -> async_solipsism.EventLoopPolicy:
-    """Return an event loop policy that uses the async solipsism event loop."""
-    return async_solipsism.EventLoopPolicy()
+@fixture(scope="session")
```
llucax: :warning:  This shouldn't be necessary and I really think we shouldn't go this way. Things should work if every test runs in its own loop and if it doesn't it means some cleanup is missing or messed up. :warning: 

Oh, I just noticed it also goes back from the correct `event_loop_policy()` to the old (and deprecated) `event_loop()` which is supposed to stop working soon :tm:. This is definitely not an option.

llucax (commented):
Very nice to see the new scheduling algorithm! :tada: Way better than what we had before.

Lots of comments, so here is a summary:

Things I think needs to be addressed before merging:

1. These commits need a commit message with a body that describes what's going on, it is really hard to review them without more information of what was wrong and how it was fixed:

    * `Fix wrong running behaviors with duration=None`
    * `Rewrite internal architecture to be more flexible`
    * `Add dispatch runner` (less important but it would be nice, is not a trivial class)
  
2. Finish the renaming of `config` -> `update`
3. The hack in the tests (session-wide `event_loop` fixture), we can't really merge this if something is so fucked up that makes tests hang or make the interpreter segfault :grimacing: 
4. Figure out if we really need to support multiple actors, it is still make noise in my head, my feeling is we should encourage having one actor to handle one dispatch type, if that actor needs other supporting actors, it should be the actor itself in charge of spawning them.
5. If we are doing more renaming (like `DispatchManagingActor` and `payload` -> `options`), it might be good to do them now too to avoid having breaking changes in the future.

The rest is minor or too big for this PR, so I will create follow-up issues.

tests/test_frequenz_dispatch.py
Marenz: I am happy to take any working alternative ;)

Marenz:
And I still think "update" sucks and "configuration" was better. "Update" doesn't fit at all in the whole naming scheme

src/frequenz/dispatch/_actor_runner.py
Marenz: how about `inputs`?

Marenz:
Addressed points 1 - 5

src/frequenz/dispatch/_managing_actor.py
```diff
@@ -0,0 +1,178 @@
...
+                    await self._updates_sender.send(
+                        DispatchUpdate(
+                            components=dispatch.selector,
+                            dry_run=dispatch.dry_run,
+                            options=dispatch.payload,
+                        )
+                    )
+
+                _logger.info("Started by dispatch %s", dispatch.id)
+                self._start_actor()
```
llucax: I just realized this is tricky. The actor needs to make sure it reads from the updates channel in the `_run()` method before doing any other work, otherwise it might be running with a missing configuration until the update message arrives. I think we should put this more clearly in the docs, including the example. But we can do it as a followup PR.

src/frequenz/dispatch/_actor_runner.py
llucax: After talking to @daniel-zullo-frequenz we agreed that in the future his use case should use only one central actor to handle dispatches that spawns more actors as needed, but that would mean a big refactoring we can start right now, so we would leave the support for multiple actors here for now. What we can do is accept `actor: Actor | Set[Actor]`, so for all normal use cases people can pass just one actor, and document the use of multiple actors is discouraged and will be removed in the future.

llucax: Oh, I see you removed the support for multiple actors already. I didn't meant we should remove it, but we should *figure out if we really need to ...* :(

src/frequenz/dispatch/_actor_runner.py
llucax: Yeah, that could work too, I think we discussed `options` at some point but wasn't perfectly accurate as some "options" might not be *optional*. I guess `inputs` is an improvement in that sense. Maybe `parameters` could work too.

src/frequenz/dispatch/_managing_actor.py
```diff
@@ -0,0 +1,178 @@
...
+        dispatcher = Dispatcher(
+            microgrid_id=microgrid_id,
+            server_url=url,
+            key=key
+        )
+
+        # Create update channel to receive dispatch update events pre-start and mid-run
+        dispatch_updates_channel = Broadcast[DispatchUpdate](name="dispatch_updates_channel")
+
+        # Start actor and supporting actor, give each a config channel receiver
```
llucax:
```suggestion
        # Start actor and supporting actor, give each an updates channel receiver
```

tests/test_mananging_actor.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+@dataclass
+class TestEnv:
+    """Test environment."""
+
+    actor: Actor
+    runner_actor: DispatchManagingActor
+    running_status_sender: Sender[Dispatch]
+    configuration_receiver: Receiver[DispatchUpdate]
```
llucax:
```suggestion
    updates_receiver: Receiver[DispatchUpdate]
```

tests/test_mananging_actor.py
```diff
@@ -0,0 +1,164 @@
...
+    generator: DispatchGenerator = DispatchGenerator()
+
+
+@fixture
+async def test_env() -> AsyncIterator[TestEnv]:
+    """Create a test environment."""
+    channel = Broadcast[Dispatch](name="dispatch ready test channel")
+    config_channel = Broadcast[DispatchUpdate](name="dispatch config test channel")
```
llucax:
```suggestion
    config_channel = Broadcast[DispatchUpdate](name="dispatch updates test channel")
```

tests/test_mananging_actor.py
```diff
@@ -0,0 +1,164 @@
...
+
+    actor = MockActor()
+
+    runner_actor = DispatchManagingActor(
+        actor=actor,
+        dispatch_type="UNIT_TEST",
+        running_status_receiver=channel.new_receiver(),
+        updates_sender=config_channel.new_sender(),
```
llucax:
```suggestion
        updates_sender=updates_channel.new_sender(),
```

tests/test_mananging_actor.py
```diff
@@ -0,0 +1,164 @@
...
+    )
+
+    runner_actor.start()
+
+    yield TestEnv(
+        actor=actor,
+        runner_actor=runner_actor,
+        running_status_sender=channel.new_sender(),
+        configuration_receiver=config_channel.new_receiver(),
```
llucax:
```suggestion
        updates_receiver=updates_channel.new_receiver(),
```

tests/test_mananging_actor.py
```diff
@@ -0,0 +1,164 @@
...
+        recurrence=replace(
+            dispatch.recurrence,
+            frequency=Frequency.UNSPECIFIED,
+        ),
+    )
+
+    await test_env.running_status_sender.send(Dispatch(dispatch))
+    fake_time.shift(timedelta(seconds=1))
+
+    event = await test_env.configuration_receiver.receive()
```
llucax:
```suggestion
    event = await test_env.updates_receiver.receive()
```

tests/test_mananging_actor.py
```diff
@@ -0,0 +1,164 @@
...
```
llucax:
```suggestion
    event = await test_env.updates_receiver.receive()
```

llucax (commented):
A few more instances of *config* and we need support for multiple actors for now (temporarily). I will also create issues for the stuff we can address in the future.

src/frequenz/dispatch/_actor_runner.py
llucax: * #55

src/frequenz/dispatch/_actor_runner.py
llucax: * #56

src/frequenz/dispatch/_managing_actor.py
llucax: * https://github.com/frequenz-floss/frequenz-dispatch-python/issues/56

Marenz:
* Added multiple actors back
* fixed confi -> up stuff

src/frequenz/dispatch/_dispatch.py
llucax: * https://github.com/frequenz-floss/frequenz-client-dispatch-python/issues/93

src/frequenz/dispatch/actor.py
llucax: * #57

src/frequenz/dispatch/actor.py
```diff
@@ -59,31 +45,99 @@ def __init__(
...
-        self._poll_timer = Timer(poll_interval, SkipMissedAndDrift())
+        self._next_event_timer = Timer(
+            timedelta(seconds=100), SkipMissedAndResync(), auto_start=False
+        )
+        """The timer to schedule the next event.
+
+        Interval is chosen arbitrarily, as it will be reset on the first event.
+        """
+
+        self._scheduled_events: list[tuple[datetime, Dispatch]] = []
```
llucax, line 64: * #58

llucax (dismissed):

llucax (dismissed):
So this should be ready for merge. @daniel-zullo-frequenz wants to do a release soon using this, so ideally we should have a release too, but I guess for that we first need to release some dependencies too, right?

We need to merge changes in the `client-dispatch` and for that some changes in the `client-base`, right? We had some changes in `channels` too but those were already released in v1.2.0, correct?

Marenz:
> [..] we first need to release some dependencies too, right?

Yes.

> We need to merge changes in the client-dispatch and for that some changes in the client-base, right? 

Almost. The changes we merged now actually recreate the streamer object so we don't need any base-client changes.
I have already released the `client-dispatch` version we need now.

> We had some changes in channels too but those were already released in v1.2.0, correct?

Correct. Those are also reflected in the current pyproject.py.


I pushed the last dependency changes and a little new feature so that the actor properly re-reads the dispatches upon disconnect now.

llucax (dismissed):

pyproject.toml
```diff
@@ -165,6 +165,7 @@ disable = [
 [tool.pytest.ini_options]
 testpaths = ["tests", "src"]
 asyncio_mode = "auto"
+asyncio_default_fixture_loop_scope = "function"
```
llucax, line 168: Maybe we should remove this and keep the default as it is?

tests/test_mananging_actor.py
```diff
@@ -0,0 +1,164 @@
...
+from frequenz.channels import Broadcast, Receiver, Sender
+from frequenz.client.dispatch.test.generator import DispatchGenerator
+from frequenz.client.dispatch.types import Frequency
+from frequenz.sdk.actor import Actor
+from pytest import fixture
+
+from frequenz.dispatch import Dispatch, DispatchManagingActor, DispatchUpdate
+
+
+@fixture(scope="session")
```
llucax: Also if it fails even with session, I would leave it to the default too here.

tests/test_frequenz_dispatch.py
```diff
@@ -15,18 +15,26 @@
...
+    DispatchEvent,
+    RunningState,
+    Updated,
+)
 from frequenz.dispatch.actor import DispatchingActor
 
 
-# This method replaces the event loop for all tests in the file.
-@fixture
+@fixture(scope="session")
```
llucax: Also if it fails even with session, I would leave it to the default too here.

pyproject.toml
Marenz: Then we get a depreciation warning saying the behavior will change and one should set a default

llucax: Oh, OK, I'm confused now. So this has nothing to do with trying to make tests pass and is just a new deprecation message in a new version of `pytest-asyncio`? If so it would be probably nice to put it in a separate commit with a corresponding commit message.

Marenz: I def. tried this to make the test pass, bit it also fixes a deprecation warning, it just didn't help though.

llucax: OK, this seems to be related to an update to https://github.com/pytest-dev/pytest-asyncio/releases/tag/v0.24.0. Issue: https://github.com/pytest-dev/pytest-asyncio/issues/924

llucax (dismissed):

llucax (approved):

llucax:
It looks like now cross-arch tests are consistently failing. This must be definitely a timing issue, the slowest the system is, the more likely we hit this bug and tests hang. Created an issue about it:

* https://github.com/frequenz-floss/frequenz-dispatch-python/issues/66

PR #62, opened 2024-09-30
Hack-Fix bug that eventloop leaks into other tests

Marenz:
This is so far the only reliable way to make the bug not happen.


llucax (commented):
I don't think this is a solution as replacing the `event_loop` fixture is deprecated IIRC, but it might be a way to dig more about why this is fixing the issue, so why we have the issue in the first place.

Marenz:
Unfortunately it also happens with this

PR #67, opened 2024-10-01
Fix overwritten fix for sdk dependency

Marenz:


Marenz:
Self-merging as this impacts users of the release probably.

PR #71, opened 2024-10-23
Fix recurrence

Marenz:
Recurrence didn't properly work as we passed in the
empty list for the by-* fiedls by default
which means "No valid time-unit",
e.g bymonth=[] means no month at all. never.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,4 +2,4 @@
 
 ## Summary
 
-* `frequenz-sdk` dependency has been extended to include version `1.0.0-rc1000`.
+This is a hot fix for recurrence not working
--- a/src/frequenz/dispatch/_dispatch.py
+++ b/src/frequenz/dispatch/_dispatch.py
@@ -222,13 +222,14 @@ class Dispatch(BaseDispatch):
             dtstart=self.start_time,
             count=count,
             until=until,
-            byminute=self.recurrence.byminutes,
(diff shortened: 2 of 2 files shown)


RELEASE_NOTES.md
```diff
@@ -2,4 +2,4 @@
 
 ## Summary
 
-* `frequenz-sdk` dependency has been extended to include version `1.0.0-rc1000`.
+This is a hot fix for recurrence not working
```
llucax, line 5: Can you describe better what the issue exactly was here too? As a user I have no idea how my code was affected before this release and what should I expect from this one.

Marenz: There is really only one user and I talked to him already ;)

Marenz: You can assume that it wasn't working at all before, the chance that it was is very low.
If it was working for the user before, they must already have a detailed understanding of this problem or they couldn't have worked around it

llucax: Well, you don't know for sure, this is a public repo. I'm sure we have plenty of external users :trollface: 

But anyway, I think it is good practice to write notes that are useful, otherwise there isn't much point in writing them at all.

llucax (approved):

RELEASE_NOTES.md
Marenz: Well, I see all the logs and request in the API, I know for sure :P

But.. yes, I suppose it is good practice.

PR #81, opened 2024-12-02
Add Dispatcher.refresh() method

Marenz:
- **Add new method Dispatcher.refresh() to fetch latest dispatches**

### DIFF
--- a/src/frequenz/dispatch/_dispatcher.py
+++ b/src/frequenz/dispatch/_dispatcher.py
@@ -213,6 +213,10 @@ class Dispatcher:
         """Start the actor."""
         self._actor.start()
 
+    async def refresh(self) -> None:
+        """Re-fetch all dispatches."""
+        await self._actor.fetch()
+
     @property
     def client(self) -> Client:
         """Return the client."""
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -14,7 +14,7 @@
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+* A new method `Dispatcher.refresh()` was added, for when you need to refresh the dispatcher's state.
 
 ## Bug Fixes
 
--- a/src/frequenz/dispatch/actor.py
+++ b/src/frequenz/dispatch/actor.py
@@ -22,6 +22,7 @@ _logger = logging.getLogger(__name__)
 """The logger for this module."""
 
 
+# pylint: disable=too-many-instance-attributes
 class DispatchingActor(Actor):
     """Dispatch actor.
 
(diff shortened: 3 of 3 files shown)


src/frequenz/dispatch/_dispatcher.py
```diff
@@ -213,6 +213,10 @@ async def start(self) -> None:
         """Start the actor."""
         self._actor.start()
 
+    async def refresh(self) -> None:
+        """Re-fetch all dispatches."""
+        await self._actor.fetch()
```
llucax, line 218: I don't think this is a good idea Design-wise. The idea is to only communicate with actors via channels once they are created. Maybe for this case is overkill, at least until we can really distribute actors in other processes/nodes, and if that is the case ok, but I would leave it as a private method and just add an ignore and a comment saying this is a hack.

llucax (commented):
Why do we need this? I guess it is only for *disaster recovery* as normally the dispatcher should be always up to date, or is the steam-based update now available yet?

src/frequenz/dispatch/_dispatcher.py
ela-kotulska-frequenz: Right thanks! My use case:
I have an actor that uses Dispacher.
When Dispacher starts (For the first time) it sends all dispaches and I receive them.

If my actor crashes and restarts latest dispaches are not send. And this is understandable because dispacher doesn't know that I restarted.
Another solution would be to create new dispacher on every restart. Would it be ok?

llucax: Just to clarify, as you commented in this review comment, I'm only referring to the call to the actor inside `refresh()`, adding `refresh()` itself is OK because `Dispatcher` is just a class that uses an actor underneath (and the class instance should always be *local*, only the actor might be in another process/node).

That said, yeah, I think if it crashes it should be OK to re-create, it shouldn't happen all the time anyway.

**But** I don't oppose to this addition, I think things could go wrong and having a way to force a refresh could still be useful in some cases. But maybe we should add some sort of `Warning` if `refresh()` is not really intended to be used in normal conditions, just to avoid confusion or people thinking they need to actively refresh the dispatcher instance.

llucax: BTW, the dispatcher should be instantiated by the app, not an actor, right? We should have only one dispatcher for the whole app/all actors, no?

Marenz: > BTW, the dispatcher should be instantiated by the app, not an actor, right? We should have only one dispatcher for the whole app/all actors, no?

good point, yes

ela-kotulska-frequenz: Ok, I will think and redesign my use case... Sorry

Marenz:
Closing in favor of https://github.com/frequenz-floss/frequenz-dispatch-python/pull/86

llucax (approved):

PR #82, opened 2024-12-02
Remove methods that were moved to client repo

Marenz:
The PR looks large, but most commits are just small changes. Tried to keep it well separated.

### DIFF
--- a/src/frequenz/dispatch/_dispatcher.py
+++ b/src/frequenz/dispatch/_dispatcher.py
@@ -255,10 +255,6 @@ class Dispatcher:
          - The payload changed
          - The dispatch was deleted
 
-        Note: Reaching the end time (start_time + duration) will not
-        send a message, except when it was reached by modifying the duration.
-
-
         Returns:
             A new receiver for dispatches whose running status changed.
         """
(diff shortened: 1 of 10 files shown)


src/frequenz/dispatch/_dispatcher.py
```diff
@@ -210,6 +210,17 @@ async def start(self) -> None:
         """Start the actor."""
         self._actor.start()
 
+    async def resend_current_running_states(self, dispatch_type: str) -> None:
```
Marenz, line 213: Maybe this? @llucax @ela-kotulska-frequenz

llucax (dismissed):
Bye bye `RunningState.DIFFERENT_TYPE`, you will not be missed :rofl: :heart:

Marenz:
> bye bye RunningState.DIFFERENT_TYPE, you will not be missed 

I missed it already, tbh. Now you have to make double-sure you check for the .type before looking at `running`. Already missed one occasion for that before that I luckily found later.

Still not convinced this is a better interface that leads to less errors.

Marenz:
Maybe the receiver should come pre-filtered for a set of types you specify :thinking:

PR #89, opened 2024-12-19
Interface enhancements

### LINKED ISSUE #87: Potentially lost dispatches on startup
### What happened?

On startup we receive all active dispatches. If there is more then 50 dispatches, they get lost.
To receive dispatches we use frequenz.channels receivers. And its default limit is 50. If there are more then 50 messages, some of them will be lost: 
Example from README.md
```py
    dispatcher = Dispatcher(
        microgrid_id=microgrid_id,
        server_url=url,
        key=key
    )
    await dispatcher.start()

    actor = MagicMock() # replace with your actor

    changed_running_status_rx = dispatcher.running_status_change.new_receiver() # HERE RECEIVERS LIMIT IS 50

    async for dispatch in changed_running_status_rx:
```

### What did you expect instead?

Find solution to receive more messages.
or  
Define maximum number of active dispatches and return receiver with correct size from dispatcher. :) 
```py
class Dispacher:
   fn running_status_change():
      return self.running_status_change.new_receiver(limit=MAX_ACTIVE_DISPATCHES)
```

### Affected version(s)

_No response_

### Affected part(s)

Hight-level interface / `Dispatcher` (part:dispatcher)

### Extra information

_No response_

Marenz:
- **Replace channel properties with receiver functions**
- **Tests: Prevent inf actor restarts on exceptions**

This also fixes #87

### DIFF
--- a/src/frequenz/dispatch/_bg_service.py
+++ b/src/frequenz/dispatch/_bg_service.py
@@ -1,8 +1,9 @@
 # License: MIT
 # Copyright © 2024 Frequenz Energy-as-a-Service GmbH
 
-"""The dispatch actor."""
+"""The dispatch background service."""
 
+import asyncio
 import logging
 from dataclasses import dataclass, field
 from datetime import datetime, timedelta, timezone
@@ -13,7 +14,7 @@ from frequenz.channels import Broadcast, Receiver, select, selected_from
 from frequenz.channels.timer import SkipMissedAndResync, Timer
 from frequenz.client.dispatch import Client
 from frequenz.client.dispatch.types import Event
-from frequenz.sdk.actor import Actor
+from frequenz.sdk.actor import BackgroundService
 
 from ._dispatch import Dispatch
 from ._event import Created, Deleted, DispatchEvent, Updated
@@ -23,13 +24,11 @@ _logger = logging.getLogger(__name__)
 
 
 # pylint: disable=too-many-instance-attributes
-class DispatchingActor(Actor):
-    """Dispatch actor.
+class DispatchScheduler(BackgroundService):
+    """Dispatch background service.
 
-    This actor is responsible for handling dispatches for a microgrid.
-
-    This means staying in sync with the API and scheduling
-    dispatches as necessary.
+    This service is responsible for managing dispatches and scheduling them
+    based on their start and stop times.
     """
 
     @dataclass(order=True)
@@ -52,7 +51,7 @@ class DispatchingActor(Actor):
         microgrid_id: int,
         client: Client,
     ) -> None:
-        """Initialize the actor.
+        """Initialize the background service.
 
         Args:
             microgrid_id: The microgrid ID to handle dispatches for.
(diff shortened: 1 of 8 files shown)


tests/test_frequenz_dispatch.py
```diff
@@ -63,24 +63,25 @@ class ActorTestEnv:
...
 
     actor = DispatchingActor(
         microgrid_id=microgrid_id,
-        lifecycle_updates_sender=lifecycle_updates_dispatches.new_sender(),
-        running_state_change_sender=running_state_change_dispatches.new_sender(),
         client=client,
     )
 
+    # pylint: disable=protected-access
+    actor._restart_limit = 0
```
llucax: Maybe we should expose these fixtures publicly in the SDK in some `test` package?

https://github.com/frequenz-floss/frequenz-sdk-python/blob/v1.x.x/tests/conftest.py

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -218,18 +212,20 @@ def client(self) -> Client:
         """Return the client."""
         return self._client
 
-    @property
-    def lifecycle_events(self) -> ReceiverFetcher[DispatchEvent]:
-        """Return new, updated or deleted dispatches receiver fetcher.
+    # pylint: disable=redefined-builtin
+    def new_lifecycle_events_receiver(self, type: str) -> Receiver[DispatchEvent]:
```
llucax: Super minor but maybe call it `dispatch_type` instead of just `type`?

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -218,18 +212,20 @@ def client(self) -> Client:
...
+        """Return new, updated or deleted dispatches receiver.
+
+        Args:
+            type: The type of the dispatch to listen for.
 
         Returns:
             A new receiver for new dispatches.
         """
-        return self._lifecycle_events_channel
+        return self._actor.new_lifecycle_events_receiver(type)
```
llucax: This is not following the actor model. You should only interact with an actor via channels, not calling methods.

I'm OK with keeping this approach though, as if we go multi-node at some point, probably having one *dispatch-cache* per mode is desirable, so all communication with the *actor* instance can be local. And also it is probably not worth adding the complexity needed to keep this as an actor at this point.

I just think we need to stop calling this an actor and inherit directly from `BackgroundService` instead, it is confusing otherwise.

llucax (commented):
Except for the *to actor or not to actor* question, LGTM.

tests/test_frequenz_dispatch.py
Marenz: Could be useful, yes

src/frequenz/dispatch/_bg_service.py
```diff
@@ -22,13 +23,12 @@
 """The logger for this module."""
 
 
-class DispatchingActor(Actor):
-    """Dispatch actor.
+# pylint: disable=too-many-instance-attributes
+class DispatchBackgroundService(BackgroundService):
```
llucax: Minor: maybe call it `DispatchScheduler`?

llucax (dismissed):

tests/test_frequenz_dispatch.py
llucax: * https://github.com/frequenz-floss/frequenz-sdk-python/issues/1136

Marenz:
Updated to lucas tiny remaining feedback, will force-merge now

PR #95, opened 2025-01-08
Redesign managing actor

Marenz:
This way we can always send a status update to a new receiver

### DIFF
--- a/src/frequenz/dispatch/_managing_actor.py
+++ b/src/frequenz/dispatch/_managing_actor.py
@@ -7,7 +7,7 @@ import logging
 from dataclasses import dataclass
 from typing import Any, Set
 
-from frequenz.channels import Receiver, Sender
+from frequenz.channels import Broadcast, Receiver
 from frequenz.client.dispatch.types import TargetComponents
 from frequenz.sdk.actor import Actor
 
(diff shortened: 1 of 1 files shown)


src/frequenz/dispatch/_managing_actor.py
```diff
@@ -106,26 +103,45 @@ async def run():
 
     def __init__(
         self,
-        actor: Actor | Set[Actor],
-        dispatch_type: str,
         running_status_receiver: Receiver[Dispatch],
-        updates_sender: Sender[DispatchUpdate] | None = None,
+        actor: Actor | Set[Actor] | None = None,
```
llucax, line 107: I was going for an actor factory. Will submit my WIP as a draft PR so maybe you can get some inspiration about it, or we can discuss pros and cons from the different approaches.

src/frequenz/dispatch/_managing_actor.py
```diff
@@ -106,26 +103,45 @@ async def run():
...
+            frozenset([actor]) if isinstance(actor, Actor) else frozenset(actor)
+        )
+
+    def new_receiver(self) -> Receiver[DispatchUpdate]:
+        """Create a new receiver for dispatch updates.
+
+        Returns:
+            A new receiver for dispatch updates.
+        """
+        return self._updates_channel.new_receiver()
```
llucax, line 144: Oh, yes! I also thought about this, but I didn't start to address this but was my next step. I think this actor should also stop being an actor and just be a background service.

I'm not sure it is needed with my approach, but I think it would be convenient anyway, I see no advantages in delegating the creation of the channel to the outside.

llucax:
* https://github.com/frequenz-floss/frequenz-dispatch-python/pull/96

Marenz:
replaced by https://github.com/frequenz-floss/frequenz-dispatch-python/pull/103

PR #96, opened 2025-01-08
WIP: Re-create actors on dispatch updates

llucax:
This PR modifies the `DispatchManagingActor` to re-create the actor when a new dispatch is received and the actor should start running instead of just start/stop the actor. To create the actor a factory function is used, which passes the initial dispatch information to the actor, so it can be properly initialized instead of having the initialization in 2 steps, the creation and the receiving of the dispatch update.

This also removes the support for multiple actors, as it was a very niche feature and makes the code more complicated.

I was also planning to make the updates sender required to simplify the code even further, it should also be very niche for someone not wanting to receive dispatch updates, almost bordering the bug.

Finally, making the `DispatchManagingActor` a background service instead of an actor and have it manage its own channel. And maybe integrating it to the dispatcher, so the user interface could potentially be as simple as:

```py
dispatcher = Dispatcher(...)
dispatcher.start_handling(
    dispatch_type,
    actor_factory=lambda initial_dispatch, dispatch_updates_receiver: MyActor.new_with_dispatch(..., initial_dispatch, dispatch_update_receiver),
    # in the future maybe also a dispatch merge strategy here
)
...
dispatcher.stop_handling(dispatch_type)
```

### DIFF
--- a/src/frequenz/dispatch/_managing_actor.py
+++ b/src/frequenz/dispatch/_managing_actor.py
@@ -4,12 +4,15 @@
 """Helper class to manage actors based on dispatches."""
 
 import logging
+from abc import abstractmethod
+from collections.abc import Callable
 from dataclasses import dataclass
-from typing import Any, Set
+from typing import Any
 
 from frequenz.channels import Receiver, Sender
 from frequenz.client.dispatch.types import TargetComponents
 from frequenz.sdk.actor import Actor
+from typing_extensions import override
 
 from ._dispatch import Dispatch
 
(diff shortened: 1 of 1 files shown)


src/frequenz/dispatch/_managing_actor.py
```diff
@@ -159,22 +208,40 @@ async def _handle_dispatch(self, dispatch: Dispatch) -> None:
             dispatch: The dispatch to handle.
         """
         if dispatch.type != self._dispatch_type:
-            _logger.debug("Ignoring dispatch %s", dispatch.id)
+            _logger.debug(
```
Marenz, line 211: We don't need this check anymore as the new running_state sender supports filtering by type

src/frequenz/dispatch/_managing_actor.py
```diff
@@ -94,58 +131,70 @@ async def run():
...
             running_status_receiver: The receiver for dispatch running status changes.
             updates_sender: The sender for dispatch events
         """
         super().__init__()
         self._dispatch_rx = running_status_receiver
-        self._actors: frozenset[Actor] = frozenset(
-            [actor] if isinstance(actor, Actor) else actor
-        )
+        self._actor_factory = actor_factory
+        self._actor: Actor | None = None
```
shsms, line 165: I thought this was going to change to `self._actors: dict[int, Actor]` where the key is int as the dispatch ID, allowing all incoming dispatches to run or `frozenset[int]` allowing only one dispatch for a given set of batteries.

The power setting actor is using int as key.  Is that coming in a separate PR @Marenz?

llucax: OK, that's a good point but a separate issue. This was a list of actors in the past only because we some use case actor needed to instantiate many actors when a dispatch arrived only because that actor was not designed properly, this is why this PR removes that.

We need to support the use case you mention too, but I would discuss it elsewhere.

llucax: (this PR might not be merged in a LONG time, while the feature you mention should be implemented soon)

Marenz:
Fyi, I started on a new branch a redesign based on this and the other or, so don't invest time in this currently

Marenz:
Also, I now accidentally pushed into this pr/branch, sorry. My PR is here: https://github.com/frequenz-floss/frequenz-dispatch-python/pull/103

llucax:
Restored. Although since you took over this I think I will just close. We can reopen if for some strange reason it would become relevant again.

PR #98, opened 2025-01-21
Implement merge strategy based on TYPE+TARGET

Marenz:


src/frequenz/dispatch/_bg_service.py
```diff
@@ -119,54 +131,87 @@ def new_lifecycle_events_receiver(self, type: str) -> Receiver[DispatchEvent]:
         )
 
     async def new_running_state_event_receiver(
-        self, type: str, *, unify_running_intervals: bool = True
+        self, type: str, *, unify_running_intervals: UnifyStrategy | None = None
```
llucax: I don't like the `Enum` design because it is not extensible and every strategy needs to be implemented inside this method, which complicates the code a lot. Can we make the strategy a class instead? This class can have the `is_still_running_filter()` function and other functions we need to change the behaviour here.

Even more, maybe the strategy can be implemented completely **outside** this method, so instead of:

```py
recv1 = dispatcher.new_running_state_event_receiver(type1, MergeByType())
recv2 = dispatcher.new_running_state_event_receiver(type2, MergeByTarget())
```

it would be:

```py
recv1 = MergeByType(dispatcher.new_running_state_event_receiver(type1))
recv2 = MergeByTarget(dispatcher.new_running_state_event_receiver(type2))
```

I didn't have a look in depth to determine if the second approach is possible or not, and even when I like the second approach being more generic, one advantage about the first approach is strategies are move visible, if you have a look at the method signature you already know you can use strategies, otherwise this could be only discovered by reading the docs.

Marenz:
~Based on https://github.com/frequenz-floss/frequenz-dispatch-python/pull/99~

src/frequenz/dispatch/_bg_service.py
```diff
@@ -23,6 +27,77 @@
...
+        scheduler: DispatchScheduler,
+    ) -> Callable[[Dispatch], bool]:
+        """Get a filter function for dispatches.
+
+        Args:
+            scheduler: The dispatch scheduler.
+
+        Returns:
+            A filter function.
+        """
```
llucax: This should all be public, an important reason to uses classes is so that it is easy to extend from outside the library itself!

llucax: Also, why `get_filter_function()` and not just `filter()`?

```suggestion
class MergeStrategy(ABC):
    """Base class for strategies to merge running intervals."""

    @abstractmethod
    def filter(self, dispatch: Dispatch) -> bool:
        """Get a filter function for dispatches.

        Args:
            scheduler: The dispatch scheduler.

        Returns:
            A filter function.
        """
```

And if any strategy needs the scheduler[^1], it should take it in the constructor.

```py
class MergeByType(_MergeStrategy):
    """Merge running intervals based on the dispatch type."""

    def __init__(self, scheduler: DispatchScheduler) -> None:
        """Initialize the strategy."""
        self._scheduler: DispatchScheduler = scheduler
```

It also feels to me that a merge strategy should be able to provide a `map()` function too (for example you might want to strip or add stuff from the `Dispatch` and return a different type), but maybe this is not really necessary :thinking: 

[^1]: Maybe we should expose the list of dispatches directly in the dispatcher, or expose the scheduler, to make creating a merge strategy more high-level.

src/frequenz/dispatch/_bg_service.py
```diff
@@ -119,54 +194,35 @@ def new_lifecycle_events_receiver(self, type: str) -> Receiver[DispatchEvent]:
...
-                    for dispatch in self._dispatches.values()
-                    if dispatch.type == type
-                )
-                # If no other dispatches are running, we can allow the stop event
-                return not other_dispatches_running
-
-            receiver = receiver.filter(_is_type_still_running)
+        if merge_strategy:
+            # pylint: disable=protected-access
+            receiver = receiver.filter(merge_strategy._get_filter_function(self))
```
llucax: And this would be:
```py
            receiver = receiver.filter(merge_strategy.filter)
```

src/frequenz/dispatch/_bg_service.py
```diff
@@ -23,6 +27,77 @@
...
+
+class MergeByType(_MergeStrategy):
+    """Merge running intervals based on the dispatch type."""
+
+    def __init__(self) -> None:
+        """Initialize the strategy."""
+        self._scheduler: DispatchScheduler
+        self._new_dispatch: Dispatch
+
+    def _get_filter_function(
```
```suggestion
    @override
    def _get_filter_function(
```

src/frequenz/dispatch/_bg_service.py
```diff
@@ -23,6 +27,77 @@
...
+        )
+        # pylint: enable=protected-access
+
+        return not other_dispatches_running
+
+
+class MergeByTypeTarget(MergeByType):
+    """Merge running intervals based on the dispatch type and target."""
+
+    def _criteria(self, dispatch: Dispatch) -> bool:
```
```suggestion
    @override
    def _criteria(self, dispatch: Dispatch) -> bool:
```

src/frequenz/dispatch/_bg_service.py
```diff
@@ -23,6 +27,77 @@
...
+    def _get_filter_function(
+        self, scheduler: DispatchScheduler
+    ) -> Callable[[Dispatch], bool]:
+        """Get a filter function for dispatches."""
+        self._scheduler = scheduler
+        return self._filter_func
+
+    def _criteria(self, dispatch: Dispatch) -> bool:
+        """Define the criteria for checking other running dispatches."""
+        return dispatch.type == self._new_dispatch.type
```
llucax: I would not save the `_new_dispatch` as an instance attribute, but pass it in the args.

```suggestion
    def _criteria(self, existing_dispatch: Dispatch, new_disaptch: Dispatch) -> bool:
        """Define the criteria for checking other running dispatches."""
        return existing_dispatch.type == new_dispatch.type
```

src/frequenz/dispatch/_bg_service.py
```diff
@@ -23,6 +27,77 @@
...
+        """
+        if new_dispatch.started:
+            return True
+
+        self._new_dispatch = new_dispatch
+
+        # pylint: disable=protected-access
+        other_dispatches_running = any(
+            dispatch.started
+            for dispatch in self._scheduler._dispatches.values()
```
llucax: `_dispatches` needs to be public too. Maybe it is best to expose the cached dispatches in the `Dispatcher` directly, as a read-only `collections.abc.Map`.

tests/test_frequenz_dispatch.py
```diff
@@ -534,14 +546,15 @@ async def test_multiple_dispatches_unify_running_intervals(
     service.start()
 
     receiver = await service.new_running_state_event_receiver(
-        "TEST_TYPE", unify_running_intervals=True
+        "TEST_TYPE", merge_strategy=MergeByType()
```
llucax: So here it would be:
```py
        "TEST_TYPE", merge_strategy=MergeByType(service)
```

If there are cases where you don't have a `Dispatcher` instance, then maybe it could take the dispatches dict directly, although it is more bug prone, if for some reason the object gets replaced or updated inside the `DispatchScheduler`, then the merge strategy will get out of sync, so I'm more inclined to get the dict every time from some other object we need is much less likely to be replaced than a `dict`. Worse case it could get a callable to get the `dict`, although is not super user-friendly, so maybe just change the test to test merge strategies in places where a `Dispatcher` instance is available.

llucax (commented):
In summary:

* I would make the `MergeStrategy` interface public and simplify it as suggested.
* Make everything needed by merge strategies available as public symbols.
* I would make merge strategies take any state they need via the constructor, not the filter function.
* Expose the cached dispatches via the `Dispatcher` directly.

src/frequenz/dispatch/_bg_service.py
Marenz: I felt, requiring the user to explicitly pass the scheduler was unnecessary work demand on them. I guess we could still do it but provide shortcuts through `Dispatcher.mergeByType()` and similar

Marenz:
Updated @llucax

Marenz:
I solved avoiding passing the service now by just using the type instead

src/frequenz/dispatch/_bg_service.py
```diff
@@ -23,6 +29,76 @@
...
+        """Initialize the strategy.
+
+        Args:
+            scheduler: The dispatch scheduler.
+        """
+        self._scheduler = scheduler
+
+    @property
+    @abstractmethod
+    def filter(self) -> Callable[[Dispatch], bool]:
```
llucax: Why is this still returning a callable and is not the callable itself? :thinking:

src/frequenz/dispatch/_bg_service.py
```diff
@@ -119,54 +200,34 @@ def new_lifecycle_events_receiver(self, type: str) -> Receiver[DispatchEvent]:
         )
 
     async def new_running_state_event_receiver(
-        self, type: str, *, unify_running_intervals: bool = True
+        self, type: str, *, merge_strategy: type[MergeStrategy] | None = None
```
llucax: I still prefer the approach of instantiating the merge strategy from outside and passing the instance for a couple of reasons:

1. Flexibility. A merge instance could potentially need to keep more state than just the scheduler, if we instantiate it inside the function we remove that flexibility of allowing merge strategies have the state they want.
2. Reuse: We could potentially reuse the same merge strategy instance with many different actors (this is minor, the above is a stronger point, but still).

If the idea is to make the typical use easier, I liked the idea of adding a wrapper for common merge strategies in `Dispatcher` for example, like `Dispatcher.mergeByWhatever()`, but I think the design/architecture of merge strategies in general should be flexible.

Then I also don't think it is much worse from the user perspective to use `MergeByWhatever(dispatcher)` than `dispatcher.mergeByWhatever()`, and it might be generally useful to have `dispatcher.dispatches` exposed publicly as read-only.

Marenz: > If the idea is to make the typical use easier, I liked the idea of adding a wrapper for common merge strategies in Dispatcher for example, like Dispatcher.mergeByWhatever(), but I think the design/architecture of merge strategies in general should be flexible.

I thought about this approach, but at the end it wouldn't save anything because then you have to type
* `dispatcher.new_running_state_event_receiver("TEST", dispatcher.mergeByExample)` instead of
* `dispatcher.new_running_state_event_receiver("TEST", MergeByExample(dispatcher))` 

So that seemed a bit useless.

Only a full wrapper like
* `dispatcher.new_rst_event_receiver_merged_by_example("TEST")` 

would make some usage difference..

llucax: Agreed. So combined to what we discussed in #104 I guess we agree that the end goal would be to end up with: `dispatcher.manage(type="FCR", actor_factory=.., strategy=MergeByType(dispatcher))`.

On the other hand... If we are merging, I guess you always need to have the full list of dispatches, otherwise how would you merge? People can still create merge strategies that require more state, but I guess the list of available dispatches is the minimum you need to do any merging on the typical case.

What about?

```py
class MergeStrategy(ABC):
    @abstractmethod
    def filter(self, all_dispatches: Map[int, Dispatch], current_dispatch: Dispatch) -> bool: ...

...
        if merge_strategy:
            receiver = receiver.filter(functools.partial(merge_strategy.filter, self._dispatches))
            # or lambda d: merge_strategy.filter(self._dispatches, d)
```

Marenz: Updated. It doesn't have the unified method `manage` yet, I intend to add that in a later PR

src/frequenz/dispatch/_bg_service.py
```diff
@@ -104,6 +125,11 @@ def __init__(
         always at index 0.
         """
 
+    @property
+    def dispatches(self) -> ValuesView[Dispatch]:
+        """All currently cached dispatches."""
+        return self._dispatches.values()
```
llucax: This seems to be unused, if that is the case I would remove it to keep the public API surface as small as possible. If it is used, why not `def dispatches(self) -> Mapping[int, Dispatch]: return self._dispatches` instead of just the values?

src/frequenz/dispatch/_bg_service.py
```diff
@@ -23,6 +28,22 @@
 """The logger for this module."""
 
 
+class MergeStrategy(ABC):
+    """Base class for strategies to merge running intervals."""
+
+    @abstractmethod
+    def filter(self, dispatches: dict[int, Dispatch], dispatch: Dispatch) -> bool:
```
llucax: I would offer a read-only view over the dispatches, as we don't want anyone to modify the dict!

```suggestion
    def filter(self, dispatches: Mapping[int, Dispatch], dispatch: Dispatch) -> bool:
```

src/frequenz/dispatch/_bg_service.py
```diff
@@ -119,54 +145,53 @@ def new_lifecycle_events_receiver(self, type: str) -> Receiver[DispatchEvent]:
...
+        are:
+
+        * [`MergeByType`][frequenz.dispatch.MergeByType] — merges all dispatches
+          of the same type
+        * [`MergeByTypeTarget`][frequenz.dispatch.MergeByTypeTarget] — merges all
+          dispatches of the same type and target
+        * [`NoMerge`][frequenz.dispatch.NoMerge] — no merging, just send all
+          events
+
+        You can make your own strategy by subclassing
```
```suggestion
        You can make your own strategy by subclassing:
```

src/frequenz/dispatch/_bg_service.py
```diff
@@ -119,54 +145,53 @@ def new_lifecycle_events_receiver(self, type: str) -> Receiver[DispatchEvent]:
         )
 
     async def new_running_state_event_receiver(
-        self, type: str, *, unify_running_intervals: bool = True
+        self, type: str, *, merge_strategy: MergeStrategy
```
llucax: I would handle the `NoMerge` case as a special case and add a `| None = None` in here. Besides having a (simpler[^1]) default, if we keep `NoMerge` as a regular strategy, we'll add a filter function to the receiver, and one that will iterate over all dispatches for no reason. IMHO the no merge case is special enough to deserve special treatment.

[^1]: I guess we can always also add `= NoMerge()` as default too.

src/frequenz/dispatch/_bg_service.py
```diff
@@ -119,54 +145,53 @@ def new_lifecycle_events_receiver(self, type: str) -> Receiver[DispatchEvent]:
...
-                other_dispatches_running = any(
-                    dispatch.started
-                    for dispatch in self._dispatches.values()
-                    if dispatch.type == type
-                )
-                # If no other dispatches are running, we can allow the stop event
-                return not other_dispatches_running
-
-            receiver = receiver.filter(_is_type_still_running)
+        if merge_strategy:
```
llucax, line 191: Currently this will always be true (well, unless some weirdo defines `__bool__` for a merge strategy, in which case it will fuck things up anyways :D). But as I said I would leave this `if` for the no-merge case.

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -200,7 +203,10 @@ def new_lifecycle_events_receiver(
         return self._bg_service.new_lifecycle_events_receiver(dispatch_type)
 
     async def new_running_state_event_receiver(
-        self, dispatch_type: str, *, unify_running_intervals: bool = True
+        self,
+        dispatch_type: str,
+        *,
+        merge_strategy: MergeStrategy = NoMerge(),
```
llucax: OK, I see you have a default in here, but again I would make it `| None = None`.

llucax (commented):
A few more comments, but looking preeeeety goooood!

src/frequenz/dispatch/_bg_service.py
```diff
@@ -119,54 +145,53 @@ def new_lifecycle_events_receiver(self, type: str) -> Receiver[DispatchEvent]:
...
         """
-        # Find all matching dispatches based on the type and collect them
         dispatches = [
             dispatch for dispatch in self._dispatches.values() if dispatch.type == type
         ]
 
-        # Create receiver with enough capacity to hold all matching dispatches
         receiver = self._running_state_status_channel.new_receiver(
-            limit=max(1, len(dispatches))
+            limit=max(30, len(dispatches))
```
llucax, line 188: This change looks sneaky. Maybe it should be split as a separate commit and/or add a comment clarifying why the change.

llucax: And I guess what we really need here is this, right?

* https://github.com/frequenz-floss/frequenz-channels-python/issues/20

Marenz: At the end I think this wasn't even the issue, but it never-the-less doesn't hurt to have a >1 minimum here

Marenz:
Updated, addressed all comments

llucax (approved):
Amazing! :tada:

PR #99, opened 2025-01-21
Fix flakey tests

Marenz:
- **Fix test_env in test that re-created it**
- **Use dispatch-client version with flakey-tests fix**


Marenz:
Depends on https://github.com/frequenz-floss/frequenz-client-dispatch-python/pull/124

llucax (approved):
I didn't exactly what was wrong and how it was fixed for the test, but I'll trust you so approving.

PR #100, opened 2025-01-24
Widen the SDK dependency range to support 1.0.0rc1500

llucax:


shsms (approved):

llucax:
Thanks @shsms, it looks like you are not a code owner, but I will just force-merge as this is trivial.

PR #103, opened 2025-01-27
Redesign actor service

Marenz:
- **Re-create actors on dispatch updates**
- **Rename DispatchManagingService to DispatchActorsService**
- **Update release notes**

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -14,6 +14,7 @@ This release introduces a more flexible and powerful mechanism for managing disp
     * It's interface has been simplified and now only requires an actor factory and a running status receiver.
     * It only supports a single actor at a time now.
     * Refer to the updated [usage example](https://frequenz-floss.github.io/frequenz-dispatch-python/latest/reference/frequenz/dispatch/#frequenz.dispatch.DispatchActorsService) for more information.
+* `DispatchUpdate` was renamed to `DispatchInfo`.
 
 ## New Features
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,7 +2,7 @@
 
 ## Summary
 
-<!-- Here goes a general summary of what this release is about -->
+This release introduces a more flexible and powerful mechanism for managing dispatch events with new strategies for merging intervals, enhanced customization options, and better overall alignment with evolving SDK dependencies. It also simplifies actor initialization while maintaining robust support for diverse dispatch scenarios.
 
 ## Upgrading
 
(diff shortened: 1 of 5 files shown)


Marenz:
~based on https://github.com/frequenz-floss/frequenz-dispatch-python/pull/98~

llucax:
Oh, damn, I did the review in #104 :-/

Can you maybe have a look at that and reply/continue the discussion here (by providing a link to the message you are replying to in the other PR so we don't lose context)?

Marenz:
review? :)

Marenz:
> Oh, damn, I did the review in https://github.com/frequenz-floss/frequenz-dispatch-python/pull/104 :-/

No you didn't. You didn't review a single commit from this PR in #104

llucax:
Mmm, I did a review of the resulting diff, so if there are no comments affecting this PR, I guess it is OK, but well, I will do a review in this PR then to double check.

Marenz:
Correction: None of your comments relates to any commit here

src/frequenz/dispatch/_actors_service.py
```diff
@@ -0,0 +1,225 @@
...
+        )
+        dispatcher.start()
+
+        status_receiver = dispatcher.new_running_state_event_receiver("EXAMPLE_TYPE")
+
+        def actor_factory(initial_dispatch, dispatch_receiver):
+            return MyActor.new_with_dispatch(initial_dispatch, dispatch_receiver)
+
+        managing_actor = DispatchManagingActor(
+            actor_factory=actor_factory,
```
llucax: In this particular case you can just use:
```suggestion
            actor_factory=MyActor.new_with_dispatch,
```

src/frequenz/dispatch/_actors_service.py
```diff
@@ -0,0 +1,225 @@
...
+from frequenz.client.dispatch.types import TargetComponents
+from frequenz.sdk.actor import Actor, BackgroundService
+
+from ._dispatch import Dispatch
+
+_logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True, kw_only=True)
+class DispatchUpdate:
```
llucax: As I said in the other PR, I would probably call this `DispatchInfo` (or maybe `DispatchDetails`?), as it will also represent the initial dispatch information, not just an "update". `initial_dispatch: DispatchUpdate` sounds a bit weird to me. `new_with_dispatch(initial_details: DispatchDetails, ...)` or `new_with_dispatch(initial_info: DispatchInfo, ...)` sounds better to me.

src/frequenz/dispatch/_actors_service.py
```diff
@@ -0,0 +1,225 @@
...
+        """Start the background service."""
+        self._tasks.add(asyncio.create_task(self._run()))
+
+    def new_receiver(self) -> Receiver[DispatchUpdate]:
+        """Create a new receiver for dispatch updates.
+
+        Returns:
+            A new receiver for dispatch updates.
+        """
+        return self._updates_channel.new_receiver()
```
llucax: Do you really need to expose this? I wouldn't add a public method for this if it is basically only used internally.

llucax (dismissed):
In the commit "Rename `DispatchManagingService` to `DispatchActorsService`" it should be `DispatchManagingActor`, right?

I would also call it something like `ActorDispatcher`, we usually don't add the suffix `Service` to stuff inheriting from `BackgroundService`, but I'm also fine with the current name too.

Other comments are also minor and mostly stylistic, so approving.

src/frequenz/dispatch/_actors_service.py
Marenz: So, `DispatchDetail` is kind of the opposite as it is an compressed / extracted dispatch.
`DispatchInfo` .. seems alright, I kind of don't like it, but I can't find any argument against it :laughing:

Marenz:
Addressed all comments, merging...

PR #104, opened 2025-01-28
Support multiple target actors for one type of dispatch

Marenz:
- **WIP: Support dipatches refering to different actors**

developed together with https://github.com/frequenz-io/frequenz-app-edge/pull/59

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -9,16 +9,16 @@ This release introduces a more flexible and powerful mechanism for managing disp
 * `Dispatcher.start` is no longer `async`. Remove `await` when calling it.
 * Two properties have been replaced by methods that require a type as parameter.
     * `Dispatcher.lifecycle_events` has been replaced by the method `Dispatcher.new_lifecycle_events_receiver(self, dispatch_type: str)`.
-    * `Dispatcher.running_status_change` has been replaced by the method `Dispatcher.new_running_state_event_receiver(self, dispatch_type: str, unify_running_intervals: bool)`.
+    * `Dispatcher.running_status_change` has been replaced by the method `Dispatcher.new_running_state_event_receiver(self, dispatch_type: str, merge_strategy: MergeStrategy)`.
 * The managing actor constructor no longer requires the `dispatch_type` parameter. Instead you're expected to pass the type to the new-receiver function.
 * The `DispatchManagingActor` class has been renamed to `DispatchActorsService`.
     * It's interface has been simplified and now only requires an actor factory and a running status receiver.
-    * It only supports a single actor at a time now.
+    * It only starts/stops a single actor at a time now instead of a set of actors.
     * Refer to the updated [usage example](https://frequenz-floss.github.io/frequenz-dispatch-python/latest/reference/frequenz/dispatch/#frequenz.dispatch.DispatchActorsService) for more information.
 * `DispatchUpdate` was renamed to `DispatchInfo`.
 
 ## New Features
 
-* A new feature "merger strategy" (`MergeByType`, `MergeByTypeTarget`) has been added to the `Dispatcher.new_running_state_event_receiver` method. Using it, you can automatically merge & unify consecutive and overlapping dispatch start/stop events of the same type. E.g. dispatch `A` starting at 10:10 and ending at 10:30 and dispatch `B` starts at 10:30 until 11:00, with the feature enabled this would in total trigger one start event, one reconfigure event at 10:30 and one stop event at 11:00.
+* A new feature "merge strategy" (`MergeByType`, `MergeByTypeTarget`) has been added to the `Dispatcher.new_running_state_event_receiver` method. Using it, you can automatically merge consecutive and overlapping dispatch start/stop events of the same type. E.g. dispatch `A` starting at 10:10 and ending at 10:30 and dispatch `B` starts at 10:30 until 11:00, with the feature enabled this would in total trigger one start event, one reconfigure event at 10:30 and one stop event at 11:00.
 * The SDK dependency was widened to allow versions up to (excluding) v1.0.0-rc1600.
-* A new function `Dispatcher.manage` was added to simplify dispatchable actor management initialization.
+* Actor management with dispatches has been simplified. Calling `Dispatcher.start_dispatching(dispatch_type, actor_factory, merge_strategy)` will begin managing your actor for the given type and merge strategy. All you need provide is an actor factory. To stop dispatching for your type, call `Dispatcher.stop_dispatching(dispatch_type)`.
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,6 +6,7 @@ This release introduces a more flexible and powerful mechanism for managing disp
 
 ## Upgrading
 
+* `Dispatcher.start` is no longer `async`. Remove `await` when calling it.
 * Two properties have been replaced by methods that require a type as parameter.
     * `Dispatcher.lifecycle_events` has been replaced by the method `Dispatcher.new_lifecycle_events_receiver(self, dispatch_type: str)`.
     * `Dispatcher.running_status_change` has been replaced by the method `Dispatcher.new_running_state_event_receiver(self, dispatch_type: str, unify_running_intervals: bool)`.
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -20,6 +20,5 @@ This release introduces a more flexible and powerful mechanism for managing disp
 ## New Features
 
 * A new feature "merger strategy" (`MergeByType`, `MergeByTypeTarget`) has been added to the `Dispatcher.new_running_state_event_receiver` method. Using it, you can automatically merge & unify consecutive and overlapping dispatch start/stop events of the same type. E.g. dispatch `A` starting at 10:10 and ending at 10:30 and dispatch `B` starts at 10:30 until 11:00, with the feature enabled this would in total trigger one start event, one reconfigure event at 10:30 and one stop event at 11:00.
-
 * The SDK dependency was widened to allow versions up to (excluding) v1.0.0-rc1600.
-
+* A new function `Dispatcher.manage` was added to simplify dispatchable actor management initialization.
--- a/src/frequenz/dispatch/_actor_dispatcher.py
+++ b/src/frequenz/dispatch/_actor_dispatcher.py
@@ -186,7 +186,8 @@ class ActorDispatcher(BackgroundService):
         else:
             _logger.info("Starting actor for dispatch type %r", dispatch.type)
             actor = self._actor_factory(
-                dispatch_update, self._updates_channel.new_receiver()
+                dispatch_update,
+                self._updates_channel.new_receiver(limit=1, warn_on_overflow=False),
             )
             self._actors[self._map_dispatch(dispatch)] = actor
 
--- a/src/frequenz/dispatch/_actor_dispatcher.py
+++ b/src/frequenz/dispatch/_actor_dispatcher.py
@@ -181,7 +181,7 @@ class ActorDispatcher(BackgroundService):
             if self._updates_sender is not None:
                 sent_str = ", sent a dispatch update instead of creating a new actor"
                 await self._updates_sender.send(dispatch_update)
-            _logger.warning(
+            _logger.info(
                 "Actor for dispatch type %r is already running%s",
                 dispatch.type,
                 sent_str,
--- a/src/frequenz/dispatch/_actor_dispatcher.py
+++ b/src/frequenz/dispatch/_actor_dispatcher.py
@@ -128,7 +128,7 @@ class ActorDispatcher(BackgroundService):
         managing_actor = ActorDispatcher(
             actor_factory=MyActor.new_with_dispatch,
             running_status_receiver=status_receiver,
-            map_dispatch=lambda dispatch: dispatch.id,
+            dispatch_identity=lambda d: d.id,
         )
 
         await run(managing_actor)
@@ -139,7 +139,7 @@ class ActorDispatcher(BackgroundService):
         self,
         actor_factory: Callable[[DispatchInfo, Receiver[DispatchInfo]], Actor],
         running_status_receiver: Receiver[Dispatch],
-        map_dispatch: Callable[[Dispatch], int],
+        dispatch_identity: Callable[[Dispatch], int],
     ) -> None:
         """Initialize the dispatch handler.
 
@@ -147,10 +147,11 @@ class ActorDispatcher(BackgroundService):
             actor_factory: A callable that creates an actor with some initial dispatch
                 information.
             running_status_receiver: The receiver for dispatch running status changes.
-            map_dispatch: A function to identify to which actor a dispatch refers.
+            dispatch_identity: A function to identify to which actor a dispatch refers.
+                By default, it uses the dispatch ID.
         """
         super().__init__()
-        self._map_dispatch = map_dispatch
+        self._dispatch_identity = dispatch_identity
         self._dispatch_rx = running_status_receiver
         self._actor_factory = actor_factory
         self._actors: dict[int, Actor] = {}
@@ -171,7 +172,7 @@ class ActorDispatcher(BackgroundService):
             options=dispatch.payload,
         )
 
-        actor: Actor | None = self._actors.get(self._map_dispatch(dispatch))
+        actor: Actor | None = self._actors.get(self._dispatch_identity(dispatch))
 
         if actor:
             sent_str = ""
@@ -189,7 +190,7 @@ class ActorDispatcher(BackgroundService):
                 dispatch_update,
                 self._updates_channel.new_receiver(limit=1, warn_on_overflow=False),
             )
-            self._actors[self._map_dispatch(dispatch)] = actor
+            self._actors[self._dispatch_identity(dispatch)] = actor
 
             actor.start()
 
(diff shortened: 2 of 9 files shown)


src/frequenz/dispatch/_bg_service.py
```diff
@@ -53,18 +53,18 @@ def filter(self) -> Callable[[Dispatch], bool]:
         """
 
 
-class MergeByType(MergeStrategy):
-    """Merge running intervals based on the dispatch type."""
+class MergeByX(MergeStrategy):
```
llucax: `MergeByIdentity`?

llucax: I even wonder if not `MergeByHash` and `identity` -> `__hash__`. After all this is what you are doing, although talking about identity might be more clear in this context. Not sure.

src/frequenz/dispatch/_bg_service.py
```diff
@@ -76,27 +76,40 @@ def _filter_func(self, new_dispatch: Dispatch) -> bool:
...
+        return hash((dispatch.type, dispatch.target))
+
+
+class MergeByTypeID(MergeByX):
+    """Merge running intervals based on the dispatch type and ID."""
+
+    @override
+    def identity(self, dispatch: Dispatch) -> int:
+        """Identity function for the merge criteria."""
+        return hash((dispatch.type, dispatch.id))
```
llucax: Why do you need to merge the `type` here? The `type` or a particular ID can´t change, right? I using just the ID as identity should be the default, i.e. there is no merging happening the identity is the ID(entity) itself.

src/frequenz/dispatch/_actors_service.py
```diff
@@ -141,18 +143,22 @@ def __init__(
         self,
         actor_factory: Callable[[DispatchUpdate, Receiver[DispatchUpdate]], Actor],
         running_status_receiver: Receiver[Dispatch],
+        map_dispatch: Callable[[Dispatch], int] = lambda dispatch: hash(dispatch.type),
```
llucax: So, for mapping by ID you would use `lambda dispatch: dispatch.id` and this will map one actor per dispatch ID, right? This seems very related to the dispatch identity too, right? Do we want to allow different combinations of this and merging?  Do we want to map and merge with different identities or should this be unified?

Also, I think we'll need some cleaning up here. When using ID as identity, more and more actors will be created. When a dispatch is done (stopped, considering recurrence) then it is done, it can't be updated to resurrect it (one need to create a new dispatch), so in that case we should somehow stop the actor mapped to that dispatch identity. I think this could be generalized, and check if there are no further scheduled dispatches for this dispatch identity when a stop signal is received, just stop and remove the actor. Even for identities like `hash(dispatch.type)` the actor will be re-created if a user creates a new dispatch, so it should work too.

src/frequenz/dispatch/_actors_service.py
```diff
@@ -0,0 +1,234 @@
...
+        """Start the background service."""
+        self._tasks.add(asyncio.create_task(self._run()))
+
+    def new_receiver(self) -> Receiver[DispatchUpdate]:
+        """Create a new receiver for dispatch updates.
+
+        Returns:
+            A new receiver for dispatch updates.
+        """
+        return self._updates_channel.new_receiver()
```
llucax: It is probably worth using `limit=1, warn_on_overflow=False` (channels v1.5.0), as we only care about the last update, right?

src/frequenz/dispatch/_actors_service.py
```diff
@@ -0,0 +1,234 @@
...
+
+    async def _start_actor(self, dispatch: Dispatch) -> None:
+        """Start all actors."""
+        dispatch_update = DispatchUpdate(
```
llucax: Nitpick: `DispatchUpdate` is starting to bother me, as we'll also now use it as the "first configuration", not really an update, so I wonder if we should rename to something like `DispatchInfo` (or just `Dispatch`, but I guess that clashes with the *raw* `Dispatch` from the client).

Also I'm thinking mid term we should probably also allow passing a dataclass (schema) for validating and loading the payload/options, same as the SDK `ConfigManager`, so the update could also be `Options | Exception | None` like config updates.

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -228,18 +231,21 @@ async def new_running_state_event_receiver(
          - The payload changed
          - The dispatch was deleted
 
-        If `unify_running_intervals` is True, running intervals from multiple
-        dispatches of the same type are considered as one continuous running
-        period. In this mode, any stop events are ignored as long as at least
-        one dispatch remains active.
+        If `unify_running_intervals` is set, running intervals from multiple
```
shsms:
```suggestion
        If a `merge_strategy` is specified, running intervals from multiple

File: src/frequenz/dispatch/_actors_service.py (hunk 3 above)
Comment (Marenz) [reply src/frequenz/dispatch/_actors_service.py]: >  This seems very related to the dispatch identity too, right? 

Yes very much so. I redesigned the previous merge approach to be only about identity because that's exactly the same function needed here. So one MergeStrategy can be used for both modules.

> Do we want to map and merge with different identities or should this be unified?

In the end, the user can pass any callback I suppose, but I would suggest we provide a wrapped experience that does both, getting the receiver and creating & passing the matching stuff to the actors service.. something like

`dispatcher.manage(type="FCR", actor_factory=.., strategy=MergeByType(dispatcher))` and it would do all the plumbing in the background.

Comment (Marenz) [reply src/frequenz/dispatch/_actors_service.py]: > Also, I think we'll need some cleaning up here. [..] in that case we should somehow stop the actor mapped to that dispatch identity [..]

But we already do that:
```        
    if actor := self._actors.pop(self._map_dispatch(stopping_dispatch), None):
            await actor.stop(msg)
```

The only trouble I see is that actors with runtime indefinite are in there forever currently

Comment (llucax) [reply src/frequenz/dispatch/_actors_service.py]: Oh, OK, I missed the `pop()` then. But this also mean that actors are not only started/stopped each time a dispatch becomes active/inactive but also the actors are destroyed while they are stopped, correct? In the current `DispatchManagingActor` the use case actors are always kept alive and only start/stopped.

I guess it is fine either way, it is just if the actor is expensive to be created, keep it created if you know it will be needed in a near future (it is scheduled in the future) might be a worthwhile optimization.

I now wonder if there could be cases where we want the actor keep running in the background even if there is no active dispatch, for example because it needs to collect data so it has some data to work with when it starts. But not for this PR, this is something we can build in the future, if needed, just brainstorming a bit.

**Comment by Marenz** [comment]
@llucax ready for review

File: src/frequenz/dispatch/_actor_dispatcher.py [hunk 7]
```diff
@@ -179,10 +185,12 @@ async def _start_actor(self, dispatch: Dispatch) -> None:
             )
         else:
             _logger.info("Starting actor for dispatch type %r", dispatch.type)
-            self._actor = self._actor_factory(
+            actor = self._actor_factory(
                 dispatch_update, self._updates_channel.new_receiver()
```
llucax: It still looks like this could benefit from `limit=1, warn_on_overflow=False`. See https://github.com/frequenz-floss/frequenz-dispatch-python/pull/104#discussion_r1933466664.

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -178,11 +186,53 @@ def __init__(
...
             self._client,
         )
+        self._actor_dispatchers: dict[str, ActorDispatcher] = {}
 
-    async def start(self) -> None:
+    def start(self) -> None:
         """Start the local dispatch service."""
         self._bg_service.start()
 
+    async def manage(
```
llucax: I would call this `start_dispatching()` to make it more explicit and also add `stop_dispatching()` as we will need that when actors are disabled and should not be dispatched anymore (this is already done in the edge app).

src/frequenz/dispatch/_actor_dispatcher.py
```diff
@@ -138,18 +139,21 @@ def __init__(
         self,
         actor_factory: Callable[[DispatchInfo, Receiver[DispatchInfo]], Actor],
         running_status_receiver: Receiver[Dispatch],
+        map_dispatch: Callable[[Dispatch], int],
```
llucax: IMHO `map_dispatch` is very opaque, I would call this `dispatch_identity` (or maybe even `dispatch_hash`, as this is what we are doing, is a hash, but maybe identity makes the particular intention of the hash here more clear).

I would also allow `None` in here too and treat mapping by `id` as a special (default) case. The identity of a dispatch is the `dispatch.id` unless you want to merge them somehow. Then you can either change the code to deal with `None` or do a `self._dispatch_identity = dispatch_identity or lambda d: d.id` at this level.

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -178,11 +186,53 @@ def __init__(
             microgrid_id,
             self._client,
         )
+        self._actor_dispatchers: dict[str, ActorDispatcher] = {}
 
-    async def start(self) -> None:
+    def start(self) -> None:
```
llucax: Unrelated to the commit?

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -178,11 +186,53 @@ def __init__(
...
+        def id_identity(dispatch: Dispatch) -> int:
+            return dispatch.id
+
+        dispatcher = ActorDispatcher(
+            actor_factory=actor_factory,
+            running_status_receiver=await self.new_running_state_event_receiver(
+                dispatch_type, merge_strategy=merge_strategy
+            ),
+            map_dispatch=(
+                id_identity if merge_strategy is None else merge_strategy.identity
```
llucax, line 264: Specially if `merge_strategy` already uses `identity`, I would just call `map_dispatch` `dispatch_identity` or even simply `identity`.

tests/test_mananging_actor.py
```diff
@@ -8,20 +8,39 @@
...
+    MergeByTypeTarget,
+    MergeStrategy,
+)
 from frequenz.dispatch._bg_service import DispatchScheduler
 
 
+@fixture
+def generator() -> DispatchGenerator:
+    """Return a dispatch generator."""
+    return DispatchGenerator()
```
llucax, line 40: For the future, we can register new fixtures to `pytest` automatically in the client by making it a [`pytest` plugin](https://docs.pytest.org/en/6.2.x/writing_plugins.html?highlight=setuptools#writing-plugins), then client users could just magically use `dispatch_generator` as a fixture without creating one or importing it, like we can do with `mocker` and other fixtures.

llucax (commented):
Just a few comments, LGTM in general, awesome to see this finally taking form!

I think it could also use some updates to the documentation, in particular getting started type of examples.

> Fix broken MergeByTypeTarge.identity function
>
> Pending: I still need to investigate why the existing tests didn't catch
this.

Is this pending for this PR? If not maybe create an issue?

src/frequenz/dispatch/_dispatcher.py
Marenz: Moved to own commit

llucax (dismissed):
Not sure if the new round was done because there was no new request for review but I had a look and for me only a documentation update is missing, but I'm also OK to leave that for a follow-up PR, so approving.

RELEASE_NOTES.md
```diff
@@ -6,6 +6,7 @@ This release introduces a more flexible and powerful mechanism for managing disp
 
 ## Upgrading
 
+* `Dispatcher.start` is no longer `async`. Remove `await` when calling it.
 * Two properties have been replaced by methods that require a type as parameter.
     * `Dispatcher.lifecycle_events` has been replaced by the method `Dispatcher.new_lifecycle_events_receiver(self, dispatch_type: str)`.
     * `Dispatcher.running_status_change` has been replaced by the method `Dispatcher.new_running_state_event_receiver(self, dispatch_type: str, unify_running_intervals: bool)`.
```
llucax: Oh, this should also be updated. No more `unify_running_intervals`. (probably do a `git grep unify_running_intervals to make sure there isn't some other hiding somewhere)

Marenz:
@llucax fixed release notes and added 2 further commits with features I found important after trying to integrate this interface in FCR & AFRR

* https://github.com/frequenz-io/frequenz-actor-fcr/pull/429
* https://github.com/frequenz-io/frequenz-actor-afrr/pull/21

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -178,11 +188,78 @@ def __init__(
...
+        dispatcher = self._actor_dispatchers.pop(dispatch_type, None)
+        if dispatcher is not None:
+            await dispatcher.stop()
+
+        if not self._actor_dispatchers:
+            self._empty_event.set()
+
+    def __await__(self) -> Generator[Any, None, bool]:
+        """Wait until all actor dispatches are stopped."""
+        return self._empty_event.wait().__await__()
```
llucax: OK, with this change now `Dispatcher` became officially a `BackgroundService` too :laughing: 

You just have to override 4 methods to get the full `BackgroundService` features (async context manager, stopping, etc.) and interface (you are already *overriding* 2 of them). Have a look at the **Background service example using custom stopping logic** example in the [docs](https://frequenz-floss.github.io/frequenz-sdk-python/v1-pre/reference/frequenz/sdk/actor/#frequenz.sdk.actor.BackgroundService).

I think to avoid making this PR last forever, this can be delayed for a future PR, but I think it makes sense to do it eventually.

src/frequenz/dispatch/_actor_dispatcher.py
```diff
@@ -178,11 +187,23 @@ async def _start_actor(self, dispatch: Dispatch) -> None:
...
+
+                actor.start()
+
+            except Exception as e:  # pylint: disable=broad-except
+                _logger.error(
+                    "Failed to start actor for dispatch type %r: %s",
+                    dispatch.type,
+                    e,
+                    exc_info=True,
+                )
```
llucax, line 206: **Not for this PR**, as it is a separate feature that can be implemented separately, but I think it might be good to do the starting in a background task and retrying forever like the edge app "dispatch subsystem" does (maybe with the new dispatcher we can even get rid of the dispatch subsystem).

Not 100% sure, as doing this is making maybe too many assumptions about what clients want, but maybe it could also be opt-in/out by passing an option like `in_background: bool, retries: int | INFINITE)`.

llucax (dismissed):
This will be an awesome release! :tada:

src/frequenz/dispatch/_dispatcher.py
Marenz: I had an error in FCR and did your suggestion in an attempt to fix it.. at the end I am not sure it was actually needed, but it's done now so :shrug: 

(the error)
```
 src/frequenz/actor/fcr/app/__main__.py:88: error: Incompatible types in "await" (actual type "Dispatcher", expected type "Awaitable[Any]")  [misc]
```

src/frequenz/dispatch/_actor_dispatcher.py
```diff
@@ -181,7 +181,7 @@ async def _start_actor(self, dispatch: Dispatch) -> None:
             if self._updates_sender is not None:
                 sent_str = ", sent a dispatch update instead of creating a new actor"
                 await self._updates_sender.send(dispatch_update)
-            _logger.warning(
+            _logger.info(
```
llucax, line 184: I didn't get this change, the commit message says " Don't warn for changing dispatch parameters: It's a normal action ", does this mean that `_start_actor` is called every time a dispatch is updated? If so, I would even make this log a debug or remove it completely, because it seems misleading, as it was never the intention to start the actor if it was just a dispatch update.

If there is a way to tell if this is called because a dispatch just started or was updated, then maybe we can log more meaningful messages.

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -183,6 +184,8 @@ def __init__(
             server_url: The URL of the dispatch service.
             key: The key to access the service.
         """
+        super().__init__(name="Dispatcher")
```
llucax: It is always better to either forward the argument or leave it empty. Creating 2 dispatcher will result in 2 instances having the same name/unique ID, which will be super confusing.

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -192,10 +195,32 @@ def __init__(
...
+    @property
+    @override
+    def is_running(self) -> bool:
+        """Whether the local dispatch service is running."""
+        return self._bg_service.is_running
+
+    @override
+    async def wait(self) -> None:
+        """Wait until all actor dispatches are stopped."""
+        await self._empty_event.wait()
```
llucax: I would `await self._bg_service` too here (probably using `gather` to do it in parallel).

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -192,10 +195,32 @@ def __init__(
...
+
+    @override
+    def cancel(self, msg: str | None = None) -> None:
+        """Stop the local dispatch service."""
+        self._bg_service.cancel(msg)
+
+        for instance in self._actor_dispatchers.values():
+            instance.cancel()
+
+        self._actor_dispatchers.clear()
```
llucax: This should be done in the `await` instead, as `stop` is `.cancel()` + `await`, so if you do it like this, stop won't really wait until all background tasks are done.

src/frequenz/dispatch/_actor_dispatcher.py
Marenz: I mean, there is a way, but it's not available to the actor_dispatcher (without restructuring).
The idea is that the dispatch instance tells us the desired state, no matter the previous state, so from that perspectives it doesn't matter whether it was an update or a new dispatch, both cases should do both, start or update a running actor instance..

llucax: I would make this a debug then, because it is completely normal and probably not very useful to get that info when things are running. But not hung to block this PR on.

llucax (approved):

PR #108, opened 2025-02-18
Some final preparations

Marenz:
- **Add method to check if a type is currently being dispatched**
- **Make actor_factory function async for more flexibility**
- **Add parameter `autostart` to `Dispatcher()` service**
- **Add retry delay and auto restart to ActorService**
- **Update Docs, Readme, Release Notes**

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,8 +21,13 @@ The [`Dispatcher` class](https://frequenz-floss.github.io/frequenz-dispatch-pyth
 
 ```python
 import os
-from frequenz.dispatch import Dispatcher
 from unittest.mock import MagicMock
+from datetime import timedelta
+
+from frequenz.dispatch import Dispatcher, DispatchInfo, MergeByType
+
+async def create_actor(dispatch: DispatchInfo, receiver: Receiver[DispatchInfo]) -> Actor:
+    return MagicMock(dispatch=dispatch, receiver=receiver)
 
 async def run():
     url = os.getenv("DISPATCH_API_URL", "grpc://fz-0004.frequenz.io:50051")
@@ -30,38 +35,19 @@ async def run():
 
     microgrid_id = 1
 
-    dispatcher = Dispatcher(
+    async with Dispatcher(
         microgrid_id=microgrid_id,
         server_url=url,
-        key=key
-    )
-    await dispatcher.start()
+        key=key,
+    ) as dispatcher:
+        await dispatcher.start_managing(
+            dispatch_type="EXAMPLE_TYPE",
+            actor_factory=create_actor,
+            merge_strategy=MergeByType(),
+            retry_interval=timedelta(seconds=10)
+        )
 
-    actor = MagicMock() # replace with your actor
-
-    changed_running_status_rx = dispatcher.new_running_state_event_receiver("MY_TYPE")
-
-    async for dispatch in changed_running_status_rx:
-        if dispatch.started:
-            print(f"Executing dispatch {dispatch.id}, due on {dispatch.start_time}")
-            if actor.is_running:
-                actor.reconfigure(
-                    components=dispatch.target,
-                    run_parameters=dispatch.payload, # custom actor parameters
-                    dry_run=dispatch.dry_run,
-                    until=dispatch.until,
-                )  # this will reconfigure the actor
-            else:
-                # this will start a new actor with the given components
-                # and run it for the duration of the dispatch
-                actor.start(
-                    components=dispatch.target,
-                    run_parameters=dispatch.payload, # custom actor parameters
-                    dry_run=dispatch.dry_run,
-                    until=dispatch.until,
-                )
-        else:
-            actor.stop()  # this will stop the actor
+        await dispatcher
 ```
 
 ## Supported Platforms
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -26,3 +26,5 @@ This release introduces a more flexible and powerful mechanism for managing disp
     * `Dispatcher.start_managing(dispatch_type, actor_factory, merge_strategy)` to manage your actor for the given type and merge strategy. All you need provide is an actor factory.
     * `Dispatcher.stop_managing(dispatch_type)` to stop dispatching for the given type.
     * `Dispatcher.is_managed(dispatch_type)` to check if dispatching is active for the given type.
+* A new method `Dispatcher.wait_for_initialization()` has been added to wait for all actors to be initialized.
+* When using `async with Dispatcher(..) as dispatcher`, the dispatcher will first wait for the dispatch service to be initialized before entering the block.
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -62,8 +62,9 @@ Further changes:
 * A new feature "merge strategy" (`MergeByType`, `MergeByTypeTarget`) has been added to the `Dispatcher.new_running_state_event_receiver` method. Using it, you can automatically merge consecutive and overlapping dispatch start/stop events of the same type. E.g. dispatch `A` starting at 10:10 and ending at 10:30 and dispatch `B` starts at 10:30 until 11:00, with the feature enabled this would in total trigger one start event, one reconfigure event at 10:30 and one stop event at 11:00.
 * The SDK dependency was widened to allow versions up to (excluding) v1.0.0-rc1800.
 * Actor management with dispatches has been simplified:
-    * `Dispatcher.start_managing(dispatch_type, actor_factory, merge_strategy)` to manage your actor for the given type and merge strategy. All you need provide is an actor factory.
+    * `Dispatcher.start_managing(dispatch_type, actor_factory, merge_strategy, retry_interval)` to manage your actor for the given type and merge strategy. All you need provide is an actor factory.
     * `Dispatcher.stop_managing(dispatch_type)` to stop dispatching for the given type.
     * `Dispatcher.is_managed(dispatch_type)` to check if dispatching is active for the given type.
+    * Dispatches that failed to start will now be retried after a delay.
 * A new method `Dispatcher.wait_for_initialization()` has been added to wait for all actors to be initialized.
 * When using `async with Dispatcher(..) as dispatcher`, the dispatcher will first wait for the dispatch service to be initialized before entering the block.
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -22,4 +22,7 @@ This release introduces a more flexible and powerful mechanism for managing disp
 
 * A new feature "merge strategy" (`MergeByType`, `MergeByTypeTarget`) has been added to the `Dispatcher.new_running_state_event_receiver` method. Using it, you can automatically merge consecutive and overlapping dispatch start/stop events of the same type. E.g. dispatch `A` starting at 10:10 and ending at 10:30 and dispatch `B` starts at 10:30 until 11:00, with the feature enabled this would in total trigger one start event, one reconfigure event at 10:30 and one stop event at 11:00.
 * The SDK dependency was widened to allow versions up to (excluding) v1.0.0-rc1800.
-* Actor management with dispatches has been simplified. Calling `Dispatcher.start_dispatching(dispatch_type, actor_factory, merge_strategy)` will begin managing your actor for the given type and merge strategy. All you need provide is an actor factory. To stop dispatching for your type, call `Dispatcher.stop_dispatching(dispatch_type)`.
+* Actor management with dispatches has been simplified:
+    * `Dispatcher.start_managing(dispatch_type, actor_factory, merge_strategy)` to manage your actor for the given type and merge strategy. All you need provide is an actor factory.
+    * `Dispatcher.stop_managing(dispatch_type)` to stop dispatching for the given type.
+    * `Dispatcher.is_managed(dispatch_type)` to check if dispatching is active for the given type.
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,11 +6,50 @@ This release introduces a more flexible and powerful mechanism for managing disp
 
 ## Upgrading
 
+A new simplified way to manage actors has been introduced:
+
+Change your code from:
+```python
+dispatcher = Dispatcher(
+    microgrid_id=microgrid_id,
+    server_url=url,
+    key=key
+)
+dispatcher.start()
+
+status_receiver = dispatcher.new_running_state_event_receiver("EXAMPLE_TYPE")
+
+managing_actor = ActorDispatcher(
+    actor_factory=MyActor.new_with_dispatch,
+    running_status_receiver=status_receiver,
+)
+
+await run(managing_actor)
+```
+
+to
+
+```python
+async with Dispatcher(
+    microgrid_id=microgrid_id,
+    server_url=url,
+    key=key
+) as dispatcher:
+    await dispatcher.start_managing(
+        dispatch_type="EXAMPLE_TYPE",
+        actor_factory=MyActor.new_with_dispatch, # now async factory!
+        merge_strategy=MergeByType,
+    )
+    await dispatcher
+```
+
+Further changes:
+
 * `Dispatcher.start` is no longer `async`. Remove `await` when calling it.
 * Two properties have been replaced by methods that require a type as parameter.
     * `Dispatcher.lifecycle_events` has been replaced by the method `Dispatcher.new_lifecycle_events_receiver(self, dispatch_type: str)`.
     * `Dispatcher.running_status_change` has been replaced by the method `Dispatcher.new_running_state_event_receiver(self, dispatch_type: str, merge_strategy: MergeStrategy)`.
-* The managing actor constructor no longer requires the `dispatch_type` parameter. Instead you're expected to pass the type to the new-receiver function.
+* The managing actor constructor no longer requires the `dispatch_type` parameter. Instead you're expected to pass the type to the new_receiver function.
 * The `DispatchManagingActor` class has been renamed to `DispatchActorsService`.
     * It's interface has been simplified and now only requires an actor factory and a running status receiver.
     * It only starts/stops a single actor at a time now instead of a set of actors.
--- a/src/frequenz/dispatch/_actor_dispatcher.py
+++ b/src/frequenz/dispatch/_actor_dispatcher.py
@@ -7,7 +7,7 @@ import asyncio
 import logging
 from collections.abc import Callable
 from dataclasses import dataclass
-from typing import Any
+from typing import Any, Awaitable
 
 from frequenz.channels import Broadcast, Receiver
 from frequenz.client.dispatch.types import TargetComponents
@@ -136,7 +136,9 @@ class ActorDispatcher(BackgroundService):
 
     def __init__(
         self,
-        actor_factory: Callable[[DispatchInfo, Receiver[DispatchInfo]], Actor],
+        actor_factory: Callable[
+            [DispatchInfo, Receiver[DispatchInfo]], Awaitable[Actor]
+        ],
         running_status_receiver: Receiver[Dispatch],
         dispatch_identity: Callable[[Dispatch], int] | None = None,
     ) -> None:
@@ -189,7 +191,7 @@ class ActorDispatcher(BackgroundService):
         else:
             try:
                 _logger.info("Starting actor for dispatch type %r", dispatch.type)
-                actor = self._actor_factory(
+                actor = await self._actor_factory(
                     dispatch_update,
                     self._updates_channel.new_receiver(limit=1, warn_on_overflow=False),
                 )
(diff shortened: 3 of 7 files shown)


src/frequenz/dispatch/_actor_dispatcher.py
```diff
@@ -159,6 +164,9 @@ def __init__(
         self._dispatch_rx = running_status_receiver
         self._actor_factory = actor_factory
         self._actors: dict[int, Actor] = {}
+        self._start_failed_dispatches: dict[int, Dispatch] = {}
```
llucax: Maybe add a comment about what this is? It doesn't look obvious at first sight. Also for this and `_actors` maybe document what the key is supposed to be too.

src/frequenz/dispatch/_actor_dispatcher.py
```diff
@@ -189,23 +198,28 @@ async def _start_actor(self, dispatch: Dispatch) -> None:
...
             except Exception as e:  # pylint: disable=broad-except
                 _logger.error(
-                    "Failed to start actor for dispatch type %r: %s",
+                    "Failed to start actor for dispatch type %r",
                     dispatch.type,
-                    e,
-                    exc_info=True,
+                    exc_info=e,
                 )
+                self._start_failed_dispatches[identity] = dispatch
```
llucax: OK, so this is another difference with how are we managing things in the app, since actors are created and removed each time a dispatch comes instead of preserving the instance and just start/stop the actor with dispatches.

I'm still not sure which approach is best, and am starting to think maybe it is worth thinking about this a bit before deciding.

src/frequenz/dispatch/_actor_dispatcher.py
```diff
@@ -222,9 +236,16 @@ async def _stop_actor(self, stopping_dispatch: Dispatch, msg: str) -> None:
             )
 
     async def _run(self) -> None:
-        """Wait for dispatches and handle them."""
-        async for dispatch in self._dispatch_rx:
-            await self._handle_dispatch(dispatch=dispatch)
+        """Run the background service."""
+        async for selected in select(self._retry_timer, self._dispatch_rx):
+            if self._retry_timer.triggered(selected) and self._start_failed_dispatches:
```
llucax: Off-topic here, but: I don't know if I'm just too used to `selected_from(selected, recv)`, but at least on this first time I see `recv.triggered(selected)` in the wild, I have to say that I find it more confusing and less obvious what it means. I find the `selected op recv` order more natural than `recv op selected`, like the `selected` is the real subject here.

FYI @shsms

README.md
```diff
@@ -33,35 +38,19 @@ async def run():
...
-        key=key
+        key=key,
+        autostart=True
+    )
+
+    dispatcher.start_dispatching(
+        dispatch_type="EXAMPLE_TYPE",
+        actor_factory=create_actor,
+        merge_strategy=MergeByType(),
+        retry_interval=timedelta(seconds=10)
```
llucax: I don't like the auto-start, now that it is a background service, I think the recommended usage should be as a context manager, I would even remove the `autostart` option.

```py
async with Dispatcher(...) as dispatcher:
    dispatcher.start_dispatching(...)
```

RELEASE_NOTES.md
```diff
@@ -6,6 +6,48 @@ This release introduces a more flexible and powerful mechanism for managing disp
...
+
+to
+
+```python
+dispatcher = Dispatcher(
+    microgrid_id=microgrid_id,
+    server_url=url,
+    key=key,
+    autostart=True
+)
```
llucax: `async with` here too.

src/frequenz/dispatch/_actor_dispatcher.py
```diff
@@ -157,6 +164,9 @@ def __init__(
         self._dispatch_rx = running_status_receiver
         self._actor_factory = actor_factory
         self._actors: dict[int, Actor] = {}
+        self._start_failed_dispatches: dict[int, Dispatch] = {}
+        self._retry_timer = Timer(retry_interval, SkipMissedAndResync())
```
llucax: The retry timer here seems a bit naive, so it's not like an actor will be retried X time after it failed but all actors will be retried on periodic X time windows.

I think it is still a reasonable approach, as implementing it properly would take more time and add more complexity with very little real-world gain, but I think this behavior should be better documented, as it is not what I would expect from a retry_interval. With this scheme it could happen that some actor is retried just 1 seconds after it first failed, which might look weird in the logs if a 60s interval was used.

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -221,26 +262,55 @@ def cancel(self, msg: str | None = None) -> None:
...
+    def is_dispatching(self, dispatch_type: str) -> bool:
+        """Check if the dispatcher is managing actors for a given dispatch type.
+
+        Args:
+            dispatch_type: The type of the dispatch to check.
+
+        Returns:
+            True if the dispatcher is managing actors for the given dispatch type.
+        """
+        return dispatch_type in self._actor_dispatchers
```
llucax, line 264: I think failed dispatches should be considered as "dispatching" too, in general when querying this you want to know if `start_dispatching` was called on a type or not, at least according to the docs "if the dispatcher is managing actors for a given dispatch type". Maybe it can be renamed to `is_managed()` to make it less ambiguous?

If necessary, we can add a way to query if there are any actors successfully created to handle the type or not.

llucax (commented):
Looks good in general but I would remove the `autostart` and change examples to use the dispatcher as a context manager. Also IMHO it needs clarification of how the retry interval works.

I would also really think about the implications of (re)creating the actor each time a dispatch arrives, just to make sure we are not missing anything important, but maybe we can delay this even for after the release, I guess it won't be such a huge impact if we change it in the short term.

src/frequenz/dispatch/_dispatcher.py
Marenz: This method does return already true even if dispatches are failing, did you just write it to make sure or because you assumed otherwise?

Anyway, renaming is fine for me

llucax: TL;DR: All good, ignore this comment.

Oh, I mixed up the `actor_dispatcher` and the `dispatcher`. If the state of which actor instances are created successfully and which have failed, then I guess we can  leave this as it is, as the same `type` could have both (successfully running and failing), so there is no one simple answer.

Marenz:
* Removed autostart
* Made examples use `async with`
* Added "wait for initial dispatch with timeout" feature
* throw when trying to get a receiver and service is not running
* Added optional persistent actor setting/parameter for actor dispatch management
* Added a proper retry system

Marenz:
* Added optional persistent actor setting/parameter for actor dispatch management

Marenz:
* Added a proper retry system

RELEASE_NOTES.md
```diff
@@ -21,4 +21,7 @@ This release introduces a more flexible and powerful mechanism for managing disp
...
 * The SDK dependency was widened to allow versions up to (excluding) v1.0.0-rc1600.
-* Actor management with dispatches has been simplified. Calling `Dispatcher.start_dispatching(dispatch_type, actor_factory, merge_strategy)` will begin managing your actor for the given type and merge strategy. All you need provide is an actor factory. To stop dispatching for your type, call `Dispatcher.stop_dispatching(dispatch_type)`.
+* Actor management with dispatches has been simplified:
+    * `Dispatcher.start_dispatching(dispatch_type, actor_factory, merge_strategy)` to manage your actor for the given type and merge strategy. All you need provide is an actor factory.
+    * `Dispatcher.stop_dispatching(dispatch_type)` to stop dispatching for the given type.
+    * `Dispatcher.is_dispatching(dispatch_type)` to check if dispatching is active for the given type.
```
Marenz: TODO need to rename this to is_managed

src/frequenz/dispatch/_actor_dispatcher.py
```diff
@@ -116,29 +118,75 @@ async def main():
...
+            """
+            return self._channel.new_receiver()
+
+        def retry(self, dispatch: Dispatch) -> None:
+            """Retry a dispatch.
+
+            Args:
+                dispatch: The dispatch information to retry.
+            """
+            asyncio.create_task(self._retry_after_delay(dispatch))
```
llucax: This is a bug, you always need to save a reference to a task, otherwise it can be garbage collected and removed from the loop.

src/frequenz/dispatch/_actor_dispatcher.py
```diff
@@ -189,21 +248,28 @@ async def _start_actor(self, dispatch: Dispatch) -> None:
...
                 actor.start()
 
             except Exception as e:  # pylint: disable=broad-except
                 _logger.error(
-                    "Failed to start actor for dispatch type %r: %s",
+                    "Failed to start actor for dispatch type %r",
                     dispatch.type,
-                    e,
-                    exc_info=True,
+                    exc_info=e,
```
llucax, line 259: You can just use `_logger.exception()` instead of explicitly passing the `exc_info`.

llucax (commented):
To be honest I'm less happy with this iteration than with the previous one.

* I don't get very well why the new `initial_fetch_timeout` is needed. Can it normally happen that it takes too long? Is it a last resort thing? I'm not sure why this has to be done at this level, and it isn't the user that should wrap the initialization in a `timeout()` if they don't want to wait for too long. Wouldn't it make more sense to provide a method like `await dispatcher.initialization_finished()`?

* I find the retry logic extremely convoluted. Why can't we just have one task with a loop like the in the edge app? Right now to try to figure out how the retry works I need to look into like 3 different places, the retry logic is very spread out.

* I don't think it is a good idea to have 2 modes now, the persistent and non-persistent mode, it also adds a lot of complexity for something we don't know we'll need yet. It was just a comment for something to look into in the future.

* Nitpick: commit message style: "Dispatches that failed to start will now be retried after a delay." ends with `.`, is not in imperative (like "Retry dispatches that failed to start after a delay"). Also what does it mean that a "dispatch failed to start". I guess is the actor that failed to start? But it is not entirely clear to me. Sometimes a one-line title is not enough. Also the commit in the previous point could answer all the questions I have there.

I really find this iteration much harder to read and to follow, and I don't see we are gaining much from the added complexity. :worried: And for me it was practically ready to be merged. Reading my comments maybe it wasn't obvious that I felt like this, so sorry about that.

I think in the future we should try to merge more quickly and iterate over the merged stuff instead, otherwise it can get a bit tiring to iterate over and over the same PR but with the code changing radically.

Marenz:
> I don't get very well why the new initial_fetch_timeout is needed. Can it normally happen that it takes too long? Is it a last resort thing? I'm not sure why this has to be done at this level, and it isn't the user that should wrap the initialization in a timeout() if they don't want to wait for too long. Wouldn't it make more sense to provide a method like await dispatcher.initialization_finished()?

That part was added after I started using the `with()` thing and noticed that it basically waits not at all till the start and immediately tries to create the receiver at which point no dispatches are available at all, so the default size of 30 (in this case) is taken for the queue size and we can't take advantage of knowing how many dispatches there are and set the receiver queue size based on that.

As a result a normal startup would give a warning for microgrids where there are >30 dispatches and I didn't like that default behavior.

> I don't think it is a good idea to have 2 modes now, the persistent and non-persistent mode, it also adds a lot of complexity for something we don't know we'll need yet. It was just a comment for something to look into in the future.

I am sorry, but it's a fact that we need it. At least the non-persistent behavior (for sahas power setting actor). And in my opinion the changes are quite minimal and non-invasive.

> I find the retry logic extremely convoluted. Why can't we just have one task with a loop like the in the edge app? Right now to try to figure out how the retry works I need to look into like 3 different places, the retry logic is very spread out.

"very spread out"' three interaction points seem not to match that statement.. and the class itself is just 300 lines, so I don't feel this puts the total complexity above acceptable either. You can't spread things a lot in 300 lines :P
Anyway, all the logic is in the extra subclass with just three interaction points, that seems better to me, not worse?


>  I don't see we are gaining much from the added complexity. 

Imo, we're gaining a lot.
* Regarding persistence: We have at least one actor type that _needs_ the non-persistent behavior. I added options so both is possible. Imo this is the best approach because we can that way evaluate easily,
everyone at their own pace. And it unblocks sahas to integrate the power_setting actor.

llucax:
> That part was added after I started using the `with()` thing and noticed that it basically waits not at all till the start and immediately tries to create the receiver at which point no dispatches are available at all, so the default size of 30 (in this case) is taken for the queue size and we can't take advantage of knowing how many dispatches there are and set the receiver queue size based on that.

OK, so I think it then can make sense to have some `async def wait_for_initialization()` or something like that, and make `__aenter__` call both `start()` and then `await wait_for_initialization()`, right?

> I am sorry, but it's a fact that we need it.

Do we? Why? Having always removing and re-creating the actor should always work, right? Re-using the actor is sort of just an optimization.

> At least the non-persistent behavior (for sahas power setting actor).

Yeah, so the mode that was previously implemented (always delete the actor when it is stopped) is the one that needs to be supported no matter what. The other mode, IMHO is not necessary, it can be implemented in the future transparently, and only delete an actor when the dispatch is completely done (it is done and have no recurrence rules, or has recurrence rules, but there are no more *next events*). In principle I don't think we need a mode to keep the actor alive always, forever. So I think we shouldn't do futurology and implement stuff we don't need yet, or don't know if it will be needed, as it adds development, maintenance and learning effort.

> And in my opinion the changes are quite minimal and non-invasive.

Maybe in terms of code, but I think it also adds complexity in terms of interface, now users need to think about this, and if we want to do a 1.0 soon(ish), it means if it was a mistake, we need to keep maintaining it.

> "very spread out"' three interaction points seem not to match that statement.. and the class itself is just 300 lines, so I don't feel this puts the total complexity above acceptable either. You can't spread things a lot in 300 lines :P
> Anyway, all the logic is in the extra subclass with just three interaction points, that seems better to me, not worse?

3 places is three times more than 1, is a 300% increase. So yeah, **very** :trollface:

Jokes aside, I still don't get why the "one simple loop" approach we have in the edge app is not sufficient here. I'm not all that worried about adding complexity, just about adding **unnecessary** complexity.

> Imo, we're gaining a lot.
> 
> * Regarding persistence: We have at least one actor type that needs the non-persistent behavior. I added options so both is possible. Imo this is the best approach because we can that way evaluate easily,
>   everyone at their own pace. And it unblocks sahas to integrate the power_setting actor.

But again, the point was never to remove "Sahas mode", so we only gained a mode that we don't even know if we need at all (and also my initial hunch is we don't, which makes me even more hesitant about adding it).

Even more, if we merged the previous iteration by just removing `autostart`, Sahas would be already unblock, but he now is still blocked. Oh dear, the Sahas dilemma :laughing: 

Anyway, I'd prefer much more to merge the previous iteration than this one, but if you feel too strongly about reverting, I can live with it, as long as the bug about the dangling task is fixed. Then we can change stuff in followup PRs.

RELEASE_NOTES.md
```diff
@@ -6,11 +6,50 @@ This release introduces a more flexible and powerful mechanism for managing disp
...
+```python
+async with Dispatcher(
+    microgrid_id=microgrid_id,
+    server_url=url,
+    key=key
+) as dispatcher:
+    await dispatcher.start_dispatching(
+        dispatch_type="EXAMPLE_TYPE",
+        actor_factory=MyActor.new_with_dispatch, # now async factory!
+        merge_strategy=MergeStrategy.MergeByType,
```
llucax: Is this right? Shouldn't it be just `MergeByType`?

RELEASE_NOTES.md
```diff
@@ -21,4 +60,10 @@ This release introduces a more flexible and powerful mechanism for managing disp
...
 * The SDK dependency was widened to allow versions up to (excluding) v1.0.0-rc1600.
-* Actor management with dispatches has been simplified. Calling `Dispatcher.start_dispatching(dispatch_type, actor_factory, merge_strategy)` will begin managing your actor for the given type and merge strategy. All you need provide is an actor factory. To stop dispatching for your type, call `Dispatcher.stop_dispatching(dispatch_type)`.
+* Actor management with dispatches has been simplified:
+    * `Dispatcher.start_dispatching(dispatch_type, actor_factory, merge_strategy, retry_interval)` to manage your actor for the given type and merge strategy. All you need provide is an actor factory.
+    * `Dispatcher.stop_dispatching(dispatch_type)` to stop dispatching for the given type.
+    * `Dispatcher.is_managed(dispatch_type)` to check if dispatching is active for the given type.
```
llucax: Optional: The asymmetry makes a bit of noise in my head now, and since *start dispatching* might be ambiguous and could be interpreted as if a proper dispatch starts, maybe `start/stop_managing()` too?

In any case, I think at least keeping a consistent wording is important to avoid confusion, so focusing either in "managing" or "dispatching" and avoid mixing the 2, so I would either say:

* `Dispatcher.start_dispatching(dispatch_type, actor_factory, merge_strategy, retry_interval)` to start dispatching a dispatch type using an actor factory and merge strategy.
* `Dispatcher.stop_dispatching(dispatch_type)` to stop dispatching for the given type.
* `Dispatcher.is_dispatching(dispatch_type)` to check if dispatching is active for the given type.

or:

* `Dispatcher.start_managing(dispatch_type, actor_factory, merge_strategy, retry_interval)` to manage dispatching for your actor for the given type using a dispatch merge strategy. All you need provide is an actor factory.
* `Dispatcher.stop_managing(dispatch_type)` to stop managing for the given dispatch type.
* `Dispatcher.is_managed(dispatch_type)` to check if the given type is being managed by this dispatcher.

(same goes for the docstrings)

src/frequenz/dispatch/_actor_dispatcher.py
```diff
@@ -116,29 +117,77 @@ async def main():
...
+            return self._channel.new_receiver()
+
+        def retry(self, dispatch: Dispatch) -> None:
+            """Retry a dispatch.
+
+            Args:
+                dispatch: The dispatch information to retry.
+            """
+            task = asyncio.create_task(self._retry_after_delay(dispatch))
+            self._tasks.add(task)
```
llucax, line 165: Now you need to finalize this object too, otherwise when the dispatcher that instantiates it is destroyed, all these tasks will be dangling until the garbage collection destroys them. My feeling is this class needs to end up being a `BackgroundService` too.

src/frequenz/dispatch/_actor_dispatcher.py
```diff
@@ -212,17 +275,33 @@ async def _stop_actor(self, stopping_dispatch: Dispatch, msg: str) -> None:
...
+        actor = self._actors.get(identity)
+
+        if actor:
             await actor.stop(msg)
+
+            del self._actors[identity]
         else:
             _logger.warning(
                 "Actor for dispatch type %r is not running", stopping_dispatch.type
             )
```
llucax, line 290: Tip:

```suggestion
        if actor := self._actors.pop(identity, None):
            await actor.stop(msg)
        else:
            _logger.warning(
                "Actor for dispatch type %r is not running", stopping_dispatch.type
            )
```

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -286,6 +348,19 @@ def client(self) -> Client:
...
+        """Enter an async context.
+
+        Start this background service.
+
+        Returns:
+            This background service.
+        """
+        self.start()
+        await self.wait_for_initialization()
+        return self
```
llucax: Just in case the parent do something extra, this should be safer:

```suggestion
        await super.__aenter__()
        await self.wait_for_initialization()
        return self
```

llucax (dismissed):
Even when there are a couple of comments that I think should probably be addressed before releasing, at this point I think I would **much rather merge as is** than to keep iterating, as each time I kind of need to review the PR entirely, when only a few parts changed. It is much easier and quicker to review a new PR instead.

RELEASE_NOTES.md
Marenz: i totally saw that coming when you suggested `is managed` instead of `is_dispatching` :laughing:

src/frequenz/dispatch/_actor_dispatcher.py
Marenz: those tasks only hang around 60 seconds or so anyway then they destroy themselves through the done callback below..

RELEASE_NOTES.md
llucax: Yeah, as I said before, the rename wasn't necessary, I just mixed up the background service and the dispatcher, and that **no change was necessary**:

* https://github.com/frequenz-floss/frequenz-dispatch-python/pull/108#discussion_r1962259101

:shrug:

src/frequenz/dispatch/_actor_dispatcher.py
llucax: I agree it is not too bad, but there could be situation on program termination that the loop can emit warnings when there are pending tasks and the loop is stopped, specially in tests. So I agree it is not urgent, but eventually I would implement proper finalization for this, as most of the time that we say "meh, who cares about finalization", it eventually comes back to bite us :D

PR #114, opened 2025-03-03
Enhancements from the last PR

Marenz:
- **Use shorter variant of the same code**
- **Clean up retry-tasks on shutdown**

### DIFF
--- a/src/frequenz/dispatch/_actor_dispatcher.py
+++ b/src/frequenz/dispatch/_actor_dispatcher.py
@@ -133,8 +133,8 @@ class ActorDispatcher(BackgroundService):
     ```
     """
 
-    class RetryFailedDispatches:
-        """Manages the retry of failed dispatches."""
+    class FailedDispatchesRetrier(BackgroundService):
+        """Manages the retring of failed dispatches."""
 
         def __init__(self, retry_interval: timedelta) -> None:
             """Initialize the retry manager.
@@ -142,10 +142,16 @@ class ActorDispatcher(BackgroundService):
             Args:
                 retry_interval: The interval between retries.
             """
+            super().__init__()
             self._retry_interval = retry_interval
             self._channel = Broadcast[Dispatch](name="retry_channel")
             self._sender = self._channel.new_sender()
-            self._tasks: set[asyncio.Task[None]] = set()
+
+        def start(self) -> None:
+            """Start the background service.
+
+            This is a no-op.
+            """
 
         def new_receiver(self) -> Receiver[Dispatch]:
             """Create a new receiver for dispatches to retry.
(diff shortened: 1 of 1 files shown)


src/frequenz/dispatch/_actor_dispatcher.py
```diff
@@ -155,6 +155,12 @@ def new_receiver(self) -> Receiver[Dispatch]:
             """
             return self._channel.new_receiver()
 
+        def cleanup(self) -> None:
+            """Clean up the retry manager."""
+            for task in self._tasks:
+                task.cancel()
+            self._tasks.clear()
```
llucax: Technically this is not enough, you also should `await` for the tasks, in case they raised an exception for example, otherwise exceptions are swallowed. This is why I think it would be best to convert this class to a proper `BackgroundService`.

src/frequenz/dispatch/_actor_dispatcher.py
```diff
@@ -295,13 +296,16 @@ async def _run(self) -> None:
...
+            try:
+                retry_recv = self._retrier.new_receiver()
+
+                async for selected in select(retry_recv, self._dispatch_rx):
+                    if retry_recv.triggered(selected):
+                        self._retrier.retry(selected.message)
+                    elif self._dispatch_rx.triggered(selected):
+                        await self._handle_dispatch(selected.message)
+            finally:
+                self._retrier.cleanup()
```
llucax: If this is a background service, you can even instantiate the retrier here inside an `async with` block (you could still keep it as an instance attribute if you need to use it in many other places, but if not I think it makes sense to instantiate it in the `run`:

```py
async def _run(self) -> None:
        """Run the background service."""
        async with RetryFailedDispatches(...) as retrier:
            ...
```

llucax (commented):
A couple more comments and a couple related thoughts:

1. Nitpick, but the class should also probably be called `FailedDispatchesRetrier`, now it is named as a function.

2. Is there any reason to make retrying optional? Is there any case you can think of where retrying is not desired? It makes the code more complicated and I don't think we really need it.

3. In the future I would like to move the [retry module](https://frequenz-floss.github.io/frequenz-client-base-python/latest/reference/frequenz/client/base/retry/) to the [core lib](https://github.com/frequenz-floss/frequenz-core-python/) so we can use it in more places, like here. If we do so, you can have a retry strategy that it is basically no retry, but the code will still be simple as the retry logic will not change.

4. Lately I was also thinking about this https://github.com/frequenz-floss/frequenz-sdk-python/discussions/1176, but it is harder to think about something generally useful for an arbitrary background service.

llucax (approved):

PR #121, opened 2025-03-27
Expose `duration` through `DispatchInfo`

### LINKED ISSUE #119: Pass `duration` in `DispatchInfo`
### What's needed?

There are dispatches that need to use the duration to decide what actions to take.  For such cases, the duration of the dispatch needs to be shared in the `DispatchInfo`.

shsms:


src/frequenz/dispatch/_actor_dispatcher.py
```diff
@@ -32,6 +32,9 @@ class DispatchInfo:
     options: dict[str, Any]
     """Additional options."""
 
+    duration: timedelta | None
+    """The duration of the dispatch."""
+
```
ela-kotulska-frequenz, line 37: Just suggestion: `duration` can be in `options` dict.
However I don't know why you need this so feel free to ignore this comment.

ela-kotulska-frequenz (approved):

src/frequenz/dispatch/_actor_dispatcher.py
```diff
@@ -229,6 +232,7 @@ async def _start_actor(self, dispatch: Dispatch) -> None:
             components=dispatch.target,
             dry_run=dispatch.dry_run,
             options=dispatch.payload,
+            duration=dispatch.duration,
```
llucax, line 235: This is tricky and can be misleading. If you use the high-level interface and you are listening to the start/stop events, or your actor is managed, this duration might not match the time your actor is really running if some dispatches were merged.

I think a better approach would be to include a list of active dispatches, and just pass the full `Dispatch` object there. Another approach would be to pass here a "merged duration" instead of the raw duration of the current dispatch. Depending on the use case one could be more useful than the other.

llucax (changes requested):
Requesting changes to avoid accidental merges until we figure this out.

This solution will not work well in the general case.

src/frequenz/dispatch/_actor_dispatcher.py
llucax: I moved the conversation to #119, I think it is more appropriate to discuss the design in the issue.

PR #122, opened 2025-03-31
Lengthen CI job timeout

Marenz:
I made it shorter once to catch flaky test that are stuck faster, but it seems the matrix job can legitimately take longer.


llucax:
Weird that this needs more than the SDK, is there really a reason why it should take so long?

llucax:
I'd say we first try with this: https://github.com/frequenz-floss/frequenz-dispatch-python/pull/123

It seems to only fail on cross-arch tests, so hopefully it doesn't happen anymore when using native arm runners.

PR #131, opened 2025-04-01
Expose src Dispatch for rare complex scenarios

Marenz:
- **Fix documentation phrasing**
- **Add `DispatchInfo._src` for when requiring more context**


shsms (approved):

llucax (approved):
LGTM as a temporary hack to unblock other work, but I think we need to find a better solution for the long term.

PR #133, opened 2025-04-04
Prepare for release v0.6.1

shsms:


shsms:
This can be 0.6.1, because the only breaking change is an increase in min patch version for `typing-extensions`, so it is very minor and not breaking.

llucax (approved):

PR #135, opened 2025-04-14
Fix that every actor instance receives every update for their type

Marenz:


llucax:
Can you maybe add a more descriptive body to the PR or the commit (or both) and the release notes? From the title I don't understand what happened and how you intend to fix it.

llucax (commented):
OK, from the code I guess what happened before is that every managed actor would receive all updates to all other managed actors, so if a dispatch handled by managed actor 1 was updated, managed actor 2 would receive that update even if it wasn't its type, or if the merge result was intended to some other actor even if it was from its managed type, right?

If so, I think the release notes should be improved to make it more clear, and to also mention this affects merged dispatches using other strategies too, right? In the extreme case of not merging at all, it even means actor handing dispatch with ID 1 was receiving updates for dispatch with ID 2, 3, etc. and now will only receive updates for dispatch with ID 1?

I think the docs should be improved too to mention how this works explicitly (probably in the method creating the new updates receiver, and in the high-level docs for `Dispatcher.start_managing()` too), otherwise actor implementors will still implement the logic to handle dispatch updates that are not intended for them.

Marenz:
> what happened before is that every managed actor would receive all updates to all other managed actors, so if a dispatch handled by managed actor 1 was updated, managed actor 2 would receive that update even if it wasn't its type, 

No no, nothing that severe. Only actors of the *same* type would receive *all* updates of the *matching* thus *same* type. This matters only for one single use case:  

When you have multiple instances per type. And so far that's only the power setting actor. And sahas is very aware as he told me about this ;)

llucax:
OK, still the release notes have to be clear, and my complete misunderstanding of the issue is a good proof of that :smile: 

Otherwise people might make the wrong assumption and think they need to do changes or whatever.

llucax (approved):

PR #138, opened 2025-04-29
Update to frequenz-client-dispatch 0.9.0

Marenz:
After this we release the fix


llucax (commented):
LGTM, but the CI thinks otherwise.

llucax:
It looks like some tests are missing the new `end_time` field.

PR #146, opened 2025-05-05
Disconnect dispatcher at cleanup

### LINKED ISSUE #145: Dispatcher doesn't disconnect with client during cleanup
### What happened?

Dispatcher doesn't disconnect with the client during cleanup (it doesn't call `await self._client.disconnect()` ).
https://github.com/frequenz-floss/frequenz-dispatch-python/blob/v0.x.x/src/frequenz/dispatch/_dispatcher.py#L216

It might cause error:
`unclosed <socket.socket fd=22, family=1, type=1, proto=0>`

### What did you expect instead?

Proper cleanup

### Affected version(s)

v0.6.2

### Affected part(s)

Hight-level interface / `Dispatcher` (part:dispatcher)

### Extra information

_No response_

Marenz:
fixes #145

### DIFF
--- a/src/frequenz/dispatch/_dispatcher.py
+++ b/src/frequenz/dispatch/_dispatcher.py
@@ -220,12 +220,13 @@ class Dispatcher(BackgroundService):
         )
         self._actor_dispatchers: dict[str, ActorDispatcher] = {}
         self._empty_event = Event()
-        self._empty_event.set()
+        self._disconnecting_future: asyncio.Future[None] | None = None
 
     @override
     def start(self) -> None:
         """Start the local dispatch service."""
         self._bg_service.start()
+        self._empty_event.set()
 
     @property
     @override
@@ -235,18 +236,22 @@ class Dispatcher(BackgroundService):
 
     @override
     async def wait(self) -> None:
-        """Wait until all actor dispatches are stopped."""
-        await asyncio.gather(self._bg_service.wait(), self._empty_event.wait())
+        """Wait until all actor dispatches are stopped and client is disconnected."""
+        if self._disconnecting_future is not None:
+            await self._disconnecting_future
 
+        await asyncio.gather(self._bg_service.wait(), self._empty_event.wait())
         self._actor_dispatchers.clear()
 
-    @override
     def cancel(self, msg: str | None = None) -> None:
-        """Stop the local dispatch service."""
+        """Stop the local dispatch service and initiate client disconnection."""
         self._bg_service.cancel(msg)
 
         for instance in self._actor_dispatchers.values():
             instance.cancel()
+
+        # Initiate client disconnection asynchronously
+        self._disconnecting_future = asyncio.ensure_future(self._client.disconnect())
 
     async def wait_for_initialization(self) -> None:
         """Wait until the background service is initialized."""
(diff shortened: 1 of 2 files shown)


src/frequenz/dispatch/_dispatcher.py
```diff
@@ -361,6 +362,25 @@ async def __aenter__(self) -> Self:
         await self.wait_for_initialization()
```
llucax, line 367:
```suggestion
        self._client.__aenter__()
        await self.wait_for_initialization()
```

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -361,6 +362,25 @@ async def __aenter__(self) -> Self:
...
+
+        Stop this background service and disconnect the client.
+
+        Args:
+            exc_type: The exception type.
+            exc_value: The exception value.
+            traceback: The traceback object.
+        """
+        await super().__aexit__(exc_type, exc_value, traceback)
+        await self._client.disconnect()
```
llucax:
```suggestion
        await self._client.__aexit__()
```

llucax (commented):
I think this is a safer/more complete/future-proof approach. Didn't think too much about the order, but I think it should be fine.

llucax (commented):
As @ela-kotulska-frequenz pointed out this also misses disconnecting at stop. If we implement it there, it should come for free to `__aexit__()` as the default implementation of `__aexit__()` for `BackgroundService` is to call `stop()` (`stop()` calls `self.cancel()` + `await self.wait()`).

For now a "simple" solution when we need to do something `async` in `cancel()` (like the call to `self._client.disconnect()` would be to store the future, like:

```py
def cancel():
    self._disconnecting_future = self._client.disconnect()

def wait():
    if self._disconnecting_future is not None:
        await self._disconnecting_future
    await super(self).wait()
```
Not super nice, as the disconnection won't really happen until the `await`, I think, but it should work for most cases. To do it right we would need to spawn a new task just to do the disconnection :-/

Marenz:
Still missing the stop() part

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -358,9 +363,27 @@ async def __aenter__(self) -> Self:
...
+        Args:
+            msg: The message to log.
+        """
+        _logger.debug("Stopping dispatcher")
+        await self._bg_service.stop(msg)
+
+        for type in self._actor_dispatchers.keys():
+            await self.stop_managing(type)
+
+        await self._client.disconnect()
```
llucax: Once you "fixed" `cancel()` and `wait()` there should be no need to override `stop()` too. In this case I guess you should move the actor stopping to `cancel()` too. And once you do that, maybe it makes sense to actually spawn a new task to do the cleanup. not sure, I guess you can also have a list of futures to wait for in the `wait()`, but you will not actually initiate the cleanup upon calling `cancel()`, only when calling `wait()`.

Marenz: Hmm so basically just remove stop() now? The actor stopping is already in cancel (using .cancel)..

llucax: Then I guess so, I hope it works. Stopping/cleanup is pretty difficult/complicated at the moment, I hope we can improve things soon.

llucax (approved):
I praise the lord that this works :pray: :laughing:

PR #155, opened 2025-05-21
Fetch dispatches in sync with periodic reconnect of client.stream

Marenz:
Should probably use
https://github.com/frequenz-floss/frequenz-client-base-python/pull/141
eventually to be less hacky?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,9 +38,11 @@ dependencies = [
   # Make sure to update the version for cross-referencing also in the
   # mkdocs.yml file when changing the version here (look for the config key
   # plugins.mkdocstrings.handlers.python.import)
-  "frequenz-sdk >= 1.0.0-rc1900, < 1.0.0-rc2100",
+  "frequenz-sdk >= 1.0.0-rc2002, < 1.0.0-rc2100",
   "frequenz-channels >= 1.6.1, < 2.0.0",
-  "frequenz-client-dispatch >= 0.10.2, < 0.11.0",
+  "frequenz-client-dispatch >= 0.11.0, < 0.12.0",
+  "frequenz-client-common >= 0.3.2, < 0.4.0",
+  "frequenz-client-base >= 0.11.0, < 0.12.0",
 ]
 dynamic = ["version"]
 
--- a/src/frequenz/dispatch/_actor_dispatcher.py
+++ b/src/frequenz/dispatch/_actor_dispatcher.py
@@ -13,6 +13,7 @@ from typing import Any, Awaitable, cast
 from frequenz.channels import Broadcast, Receiver, Sender, select
 from frequenz.channels.timer import SkipMissedAndDrift, Timer
 from frequenz.client.common.microgrid.components import ComponentCategory, ComponentId
+from frequenz.client.dispatch.types import TargetComponents as ClientTargetComponents
 from frequenz.sdk.actor import Actor, BackgroundService
 
 from ._dispatch import Dispatch
@@ -190,21 +191,10 @@ class ActorDispatcher(BackgroundService):
         """Start the background service."""
         self._tasks.add(asyncio.create_task(self._run()))
 
-    def _get_target_components_from_dispatch(
-        self, dispatch: Dispatch
-    ) -> TargetComponents:
-        if all(isinstance(comp, int) for comp in dispatch.target):
-            # We've confirmed all elements are integers, so we can cast.
-            int_components = cast(list[int], dispatch.target)
-            return [ComponentId(cid) for cid in int_components]
-        # If not all are ints, then it must be a list of ComponentCategory
-        # based on the definition of ClientTargetComponents.
-        return cast(list[ComponentCategory], dispatch.target)
-
     async def _start_actor(self, dispatch: Dispatch) -> None:
         """Start the actor the given dispatch refers to."""
         dispatch_update = DispatchInfo(
-            components=self._get_target_components_from_dispatch(dispatch),
+            components=_convert_target_components(dispatch.target),
             dry_run=dispatch.dry_run,
             options=dispatch.payload,
             _src=dispatch,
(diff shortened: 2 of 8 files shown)


src/frequenz/dispatch/_bg_service.py
```diff
@@ -132,6 +132,9 @@ def __init__(
         self._initial_fetch_event = asyncio.Event()
         """The initial fetch event."""
 
+        self._last_stream_event = datetime.now(timezone.utc)
```
llucax: Wouldn't it make more sense to add an option to the grpc streamer to [pass a `timeout` to the call](https://grpc.github.io/grpc/python/grpc_asyncio.html#grpc.aio.UnaryStreamMultiCallable)? Not sure if that timeout could already cause a re-connection, or we can still call the new `reconnect()` as needed.

llucax: The `timeout` can be passed at the dispatch API level, as we pass a callable that will call the gRPC method to the broadcaster, so that lambda can add the `timeout`. We already do this in the [microgrid API client](https://github.com/frequenz-floss/frequenz-client-microgrid-python/blob/d2728c913306320b1c9c68b843b773133feee74a/src/frequenz/client/microgrid/_client.py#L158) (for non-streaming methods).

But we need to check if the gRPC broadcaster will retry if a timeout exception is raised.

Marenz:
Blocked by https://github.com/frequenz-floss/frequenz-client-base-python/pull/146

ela-kotulska-frequenz (commented):
LGTM, only needs rebasing :)

Marenz:
> LGTM, only needs rebasing :)

Unfortunately, using a new base-client also means allowing a new base client version in microclient, so we will have to release both base-client and microclient

llucax:
"microclient" :laughing:

src/frequenz/dispatch/_bg_service.py
```diff
@@ -244,24 +246,40 @@ async def _run(self) -> None:
...
-                        self._dispatches[dispatch.id] = dispatch
-                        await self._lifecycle_events_tx.send(Updated(dispatch=dispatch))
-                    case Event.DELETED:
-                        self._dispatches.pop(dispatch.id)
-                        await self._update_dispatch_schedule_and_notify(None, dispatch)
-
-                        await self._lifecycle_events_tx.send(Deleted(dispatch=dispatch))
+                match selected.message:
+                    case StreamStartedEvent():
+                        _logger.info("Dispatch stream restarted, refreshing dispatches")
```
llucax: "(re)started"?

Marenz: Realistically we never will get the start notification because we create the `new_receiver()` only after it was sent

llucax: Unless the channel's `resend_latest` is `True` and there are no messages in between. But point taken.

Marenz: I think I'll actually will set it to resend_latest=true, it makes for a more consistent experience and also simplifies some parts on the dispatch side

llucax: Really? What can it simplify? In the general case it will only give you the last dispatch update received in the channel,, you'll always need to go get a fresh list of dispatches anyway, receiving the latest one doesn't really bring much, does it?

pyproject.toml
```diff
@@ -38,9 +38,12 @@ dependencies = [
   # Make sure to update the version for cross-referencing also in the
   # mkdocs.yml file when changing the version here (look for the config key
   # plugins.mkdocstrings.handlers.python.import)
-  "frequenz-sdk >= 1.0.0-rc1900, < 1.0.0-rc2100",
+  #"frequenz-sdk >= 1.0.0-rc1900, < 1.0.0-rc2100",
+  "frequenz-sdk @ git+https://github.com/frequenz-floss/frequenz-sdk-python.git@refs/pull/1230/head",
   "frequenz-channels >= 1.6.1, < 2.0.0",
-  "frequenz-client-dispatch >= 0.10.1, < 0.11.0",
+  #  "frequenz-client-dispatch >= 0.10.2, < 0.11.0",
+  "frequenz-client-dispatch @ git+https://github.com/frequenz-floss/frequenz-client-dispatch-python.git@refs/pull/180/head",
```
llucax: You should depend on `client-common` too.

src/frequenz/dispatch/_actor_dispatcher.py
```diff
@@ -337,3 +327,13 @@ async def _handle_dispatch(self, dispatch: Dispatch) -> None:
...
+
+
+def _convert_target_components(target: ClientTargetComponents) -> TargetComponents:
+    if all(isinstance(comp, int) for comp in target):
+        # We've confirmed all elements are integers, so we can cast.
+        int_components = cast(list[int], target)
+        return [ComponentId(cid) for cid in int_components]
+    # If not all are ints, then it must be a list of ComponentCategory
+    # based on the definition of ClientTargetComponents.
+    return cast(list[ComponentCategory], target)
```
llucax, line 317: I don ´t get this, wasn't the whole point of the new target classes to enable using `match` more easily? Why do you have TargetComponents = list[ComponentId] | list[ComponentCategory]` here which makes distinguishing between the different ways to set targets harder? Shouldn't you just use the same `TargetComponents` defined by the client here so end users can just do:

```py
match target:
    case TargetIds() as ids:
        ...
    case TargetICategories() as cats:
        ...
```
?

If you really need to keep the local `TargetComponents`, can't this function be implemented using `match`?

```py
match target:
    case TargetIds() as ids:
        return list(ids)
    case TargetICategories() as cats:
        return list(cats)
```

Marenz: I haven't really looked at the function (nor wrote it), nor do I care much about it as its just temporary until the dispatch-client supports base ids.

I just needed it accessible and modular to use it in tests.

llucax: OK, I was just surprised that you decided to rewrite it but chose not to use the *new way*.

Marenz: But I didn't? I really just moved it out of the class.

Marenz:
This is now ready

src/frequenz/dispatch/_bg_service.py
Marenz: Found a different approach now

llucax (approved):

PR #159, opened 2025-06-02
Update release notes

ela-kotulska-frequenz:


llucax (commented):
Is this updating the release notes (as cleaning them up **BEFORE** release) or resetting them (as cleaning them up **AFTER** the release), because there are 2 commits doing both :thinking:

ela-kotulska-frequenz:
> Is this updating the release notes (as cleaning them up BEFORE release) or resetting them (as cleaning them up AFTER the release), because there are 2 commits doing both 🤔

It is intentional.
I plan to checkout to the`update release notes` commit and tag it. That way I won't forget to reset notes at the end. And it is less work.

llucax:
But it is more error prone. In any case, the solution here is to automate the cleanup of the release notes (both before and after the release). Some day... some day... :)

PR #162, opened 2025-06-05
Fixed that dispatches are never retried, but forever logged.

Marenz:


src/frequenz/dispatch/_actor_dispatcher.py
```diff
@@ -323,7 +323,14 @@ async def _run(self) -> None:
 
             async for selected in select(retry_recv, self._dispatch_rx):
                 if retry_recv.triggered(selected):
-                    self._retrier.retry(selected.message)
+                    dispatch = selected.message
+                    if not dispatch.started:
+                        _logger.info(
+                            "Giving up on dispatch %s, duration was exceeded.",
+                            dispatch.id,
+                        )
```
ela-kotulska-frequenz: What `duration`? 
From the code you try once, there is no duration. But maybe I don't see something?

ela-kotulska-frequenz (commented):
LGTM just one comment

src/frequenz/dispatch/_actor_dispatcher.py
```diff
@@ -318,21 +266,35 @@ async def _stop_actor(self, stopping_dispatch: Dispatch, msg: str) -> None:
...
+                if not self._pending_dispatches:
+                    continue
+
+                _logger.info(
+                    "Retrying %d pending dispatches",
+                    len(self._pending_dispatches),
+                )
+                keys = list(self._pending_dispatches.keys())
+                for identity in keys:
+                    dispatch = self._pending_dispatches[identity]
```
ela-kotulska-frequenz: Could you please remove dispatch from `self._pending_dispatches` here, not in `self._handle_dispatch`.
1. It will be easier to read and more straightforward
2. Removing dispatch from `_pending_dispatches`, at the beginning of method called `handle_dispatch` is confusing :D

ela-kotulska-frequenz (commented):
Is it possible that you retry dispatch that just started?
Is that correct approach?

src/frequenz/dispatch/_actor_dispatcher.py
Marenz: I can also remove it here, but it is part of the design that handle dispatch removes it because if an updated dispatch instruction comes in, we need to erase the old pending one to not try to repeat with old outdated dispatches

Marenz:
> Is it possible that you retry dispatch that just started?

Yes of course, I want to only retry dispatches that started. 
`Dispatch.started` does not mean the actor is also started, only that the dispatch wants it to be started.

src/frequenz/dispatch/_actor_dispatcher.py
```diff
@@ -225,18 +173,18 @@ def __init__(  # pylint: disable=too-many-arguments, too-many-positional-argumen
...
             dispatch_identity if dispatch_identity else lambda d: d.id
         )
 
         self._dispatch_rx = running_status_receiver
+        self._retry_timer_rx = Timer(retry_interval, SkipMissedAndDrift())
         self._actor_factory = actor_factory
         self._actors: dict[int, ActorDispatcher.ActorAndChannel] = {}
-
-        self._retrier = ActorDispatcher.FailedDispatchesRetrier(retry_interval)
+        self._pending_dispatches: dict[int, Dispatch] = {}
```
ela-kotulska-frequenz: Oww I god it!
Than Maybe rename it to `_failed_dispatches`? `pending` feels like they just started and didn't finished, yet.

src/frequenz/dispatch/_actor_dispatcher.py
```diff
@@ -142,59 +143,6 @@ async def main():
     ```
     """
 
-    class FailedDispatchesRetrier(BackgroundService):
```
llucax, line 145: So satisfying to see this class go away... :rofl:

src/frequenz/dispatch/_actor_dispatcher.py
```diff
@@ -318,21 +266,35 @@ async def _stop_actor(self, stopping_dispatch: Dispatch, msg: str) -> None:
...
 
     async def _handle_dispatch(self, dispatch: Dispatch) -> None:
         """Handle a dispatch.
 
+        Removes the dispatch from the failed dispatches cache
+
         Args:
             dispatch: The dispatch to handle.
         """
+        identity = self._dispatch_identity(dispatch)
```
llucax, line 297: Maybe it would be good to put a comment here about what you replied to Ela (or in the docstring), I agree at first sight it might look not super obvious wht removing dispatches from the pending list here.

llucax (commented):
Also LGTM.

PR #164, opened 2025-06-12
Pass through parameters to client

Marenz:
- **Pass through timeout arguments for the client**
- **Harden potentially flakey test**
- **Update release notes**

### DIFF
--- a/src/frequenz/dispatch/_dispatcher.py
+++ b/src/frequenz/dispatch/_dispatcher.py
@@ -198,12 +198,15 @@ class Dispatcher(BackgroundService):
         ```
     """
 
+    # pylint: disable-next=too-many-arguments
     def __init__(
         self,
         *,
         microgrid_id: int,
         server_url: str,
         key: str,
+        call_timeout: timedelta = timedelta(seconds=60),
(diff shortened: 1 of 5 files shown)


src/frequenz/dispatch/_dispatcher.py
```diff
@@ -198,23 +198,33 @@ async def run():
         ```
     """
 
+    # pylint: disable-next=too-many-arguments
     def __init__(
         self,
         *,
         microgrid_id: int,
```
llucax, line 205: Nothing to do with this PR but I wonder if there is a reason why you need to pass the `microgrid_id` when creating the client. For edge clients this makes sense, but for cloud clients? You need to create one client per microgrid? This feels weird.

llucax (commented):
What copilot said, other than that, LGTM.

src/frequenz/dispatch/_dispatcher.py
Marenz: .. this is an edge client?!

Marenz: And, well. We need to know the id we want dispatches for, that's why we need the microgrid id.

PR #165, opened 2025-06-12
Properly clean up timer in `stop()`

Marenz:


Marenz:
Interesting that the CI fails when I add the stop method (one test)...

src/frequenz/dispatch/_bg_service.py
```diff
@@ -219,6 +220,12 @@ async def new_running_state_event_receiver(
 
     # pylint: enable=redefined-builtin
 
+    @override
+    async def stop(self, msg: str | None = None) -> None:
+        """Stop the background service."""
+        self._next_event_timer.stop()
```
llucax: We are probably going to deprecate `.stop()` as it was something specific for `Timer`, but now we have `close()` in the `Receiver` interface, so it might be more future-proof to use `close()`.


Also since this is a background service, it is better to put the `close()` in `cancel()` instead of `stop()`, otherwise if you `d.cancel(); await d` the timer won't be closed.

Finally this might be more complicated than you thought, at least we had issues in some other places where now you start getting a `ReceiverStoppedError`/`StopAsyncIteration` exception in the `run` you might need to start catching. Maybe this is a more actor-specific issue because of the auto-restart and is not really an issue for pure background services though, but just something to have in mind.

Marenz:
Changed to `close()` and added `reset()` in start to fix failing test and also because it's correct

llucax (commented):
I'm starting to think what's really more correct is:

```py
    async def _run(self) -> None:
        with closing(Timer(
            timedelta(seconds=100), SkipMissedAndResync()
        ) as next_event_timer:
            ...
```

This way you don't need to mess with start/stop/cancel, you don't need `auto_start=False`, and you have guaranteed cleanup. Do you see any problems with this?

BTW, probably the timer's period should be a constructor argument with 100 seconds just as a default.

Marenz:
> BTW, probably the timer's period should be a constructor argument with 100 seconds just as a default.

Why? It's overwritten in `update_timer()` for pretty much any activity

llucax:
We talked already via DM but just for the records, it is basically to ensure proper cleanup, and to tie the lifetime of the timer to the lifetime of the task/run. Maybe is not worth it as we have other instance variables that would also make sense to move to run, but that is too much work and if only do it for the timer, we might end up in the middle of 2 worlds, so I leave it up to you.

src/frequenz/dispatch/_bg_service.py
```diff
@@ -191,6 +185,10 @@ async def new_running_state_event_receiver(
         if not self._tasks:
             raise RuntimeError("Dispatch service not started")
 
+        (task,) = self._tasks
+        if task.done():
```
llucax: `self.is_running`?

llucax (commented):
Not approving because of the comment and auto-merge but feel free to self-merge if you address it.

src/frequenz/dispatch/_bg_service.py
Marenz: I swear, I was looking for it

PR #169, opened 2025-06-26
Bug fixes after testing with edge app

Marenz:
- **Fix sdk depency after yanked release**
- **Explictly require newer time when comparing in fetch()**
- **Harden sync for delete events between stream/fetch**
- **Harden stream/fetch sync for CREATED and UPDATED**

### DIFF
--- a/src/frequenz/dispatch/_bg_service.py
+++ b/src/frequenz/dispatch/_bg_service.py
@@ -359,7 +359,7 @@ class DispatchScheduler(BackgroundService):
                             dispatch, None, timer
                         )
                         await self._lifecycle_events_tx.send(Created(dispatch=dispatch))
-                    elif dispatch.update_time != old_dispatch.update_time:
+                    elif dispatch.update_time > old_dispatch.update_time:
                         _logger.debug("Updated dispatch: %s", dispatch)
                         await self._update_dispatch_schedule_and_notify(
                             dispatch, old_dispatch, timer
--- a/src/frequenz/dispatch/_bg_service.py
+++ b/src/frequenz/dispatch/_bg_service.py
@@ -269,11 +269,12 @@ class DispatchScheduler(BackgroundService):
                             _logger.debug(
                                 "Received dispatch event: %s", selected.message
                             )
-                            dispatch = Dispatch(selected.message.dispatch)
-                            _existing_dispatch = self._dispatches.get(dispatch.id)
+                            new_dispatch = Dispatch(selected.message.dispatch)
+                            _existing_dispatch = self._dispatches.get(new_dispatch.id)
                             is_new_or_newer = (
                                 _existing_dispatch is None
-                                or dispatch.update_time > _existing_dispatch.update_time
+                                or new_dispatch.update_time
+                                > _existing_dispatch.update_time
                             )
 
                             match selected.message.event:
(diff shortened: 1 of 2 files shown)


src/frequenz/dispatch/_bg_service.py
```diff
@@ -263,6 +264,17 @@ async def _run(self) -> None:
...
+                    def is_more_relevant(
+                        dispatch: Dispatch,
+                    ) -> bool:
+                        existing_dispatch = self._dispatches.get(dispatch.id)
+
+                        return (
+                            not existing_dispatch
+                            or dispatch.update_time > existing_dispatch.update_time
+                        )
+
```
llucax: Since you are calculating this for every case, why don't you just store it in a variable instead of creating a nested function? Also, `is_more_relevant()` seems a bit too vague, what about `is_new_or_newer`?

src/frequenz/dispatch/_bg_service.py
```diff
@@ -271,25 +283,41 @@ async def _run(self) -> None:
                             dispatch = Dispatch(selected.message.dispatch)
```
llucax: It might also add some clarity to rename `dispatch` to `new_dispatch` and use `old_dispatch` for the one we have stored, although is out of the scope of this PR, but mentioning it because it related.

I also strong recommend to use `is None` instead of `not`, as if in the future someone adds a `__bool__` to dispatch, the meaning could change. In general is better practice to use `is None` than implicit conversion to `bool` if you want to check if something returned `None`, we already had several bugs because we used `not x` with `int`/`float` when the value was `0`.

```suggestion
                            new_dispatch = Dispatch(selected.message.dispatch)
                            old_dispatch = self._dispatches.get(new_dispatch.id)
                            is_new_or_newer = (
                                old_dispatch is None
                                or new_dispatch.update_time > old_dispatch.update_time
                            )

```

src/frequenz/dispatch/_bg_service.py
```diff
@@ -271,25 +283,41 @@ async def _run(self) -> None:
...
                             match selected.message.event:
                                 case Event.CREATED:
-                                    self._dispatches[dispatch.id] = dispatch
-                                    await self._update_dispatch_schedule_and_notify(
-                                        dispatch, None, next_event_timer
-                                    )
+                                    # Check if the dispatch already exists and
+                                    # was updated. The CREATE event is late in
+                                    # this case
+                                    if is_more_relevant(dispatch):
```
```suggestion
                                    if is_new_or_newer:
```

etc.

src/frequenz/dispatch/_bg_service.py
```diff
@@ -359,7 +393,7 @@ async def _fetch(self, timer: Timer) -> None:
                             dispatch, None, timer
                         )
                         await self._lifecycle_events_tx.send(Created(dispatch=dispatch))
-                    elif dispatch.update_time != old_dispatch.update_time:
+                    elif dispatch.update_time > old_dispatch.update_time:
```
llucax, line 392: > Explictly require newer time when comparing in fetch()

It would be good to specify in the commit message **why** this change is needed or better. Maybe it is because of the limited diff context, but it is not obvious for me.

llucax (dismissed):
All optional (although I strongly recommend avoiding `not x` when you really mean `x is None`).

Disabling auto-merge to give you the opportunity to address some of the feedback.

PR #174, opened 2025-07-16
Fix dispatch dict passed to `MergeStrategy` not getting updated for future fetch's

Marenz:
* We are also switching to use the same `now` while evaluating other dispatches in the MergeStrategy, but this is just an enhancement and not a fix
* We still used logging without "path" in some places, that was fixed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,7 @@ dependencies = [
   # plugins.mkdocstrings.handlers.python.import)
   "frequenz-sdk >= 1.0.0-rc2100, < 1.0.0-rc2200",
   "frequenz-channels >= 1.6.1, < 2.0.0",
-  "frequenz-client-dispatch >= 0.11.0, < 0.12.0",
+  "frequenz-client-dispatch >= 0.11.1, < 0.12.0",
   "frequenz-client-common >= 0.3.2, < 0.4.0",
   "frequenz-client-base >= 0.11.0, < 0.12.0",
 ]
--- a/src/frequenz/dispatch/_bg_service.py
+++ b/src/frequenz/dispatch/_bg_service.py
@@ -337,7 +337,7 @@ class DispatchScheduler(BackgroundService):
                             pass
 
     async def _execute_scheduled_event(self, dispatch: Dispatch, timer: Timer) -> None:
-        """Execute a scheduled event.
+        """Execute a scheduled event and schedules the next one if any.
 
         Args:
             dispatch: The dispatch to execute.
--- a/src/frequenz/dispatch/_bg_service.py
+++ b/src/frequenz/dispatch/_bg_service.py
@@ -366,8 +366,7 @@ class DispatchScheduler(BackgroundService):
         """
         self._initial_fetch_event.clear()
 
-        old_dispatches = self._dispatches
-        self._dispatches = {}
+        new_dispatches = {}
 
         try:
             _logger.debug("Fetching dispatches for microgrid %s", self._microgrid_id)
@@ -381,9 +380,9 @@ class DispatchScheduler(BackgroundService):
                         continue
                     dispatch = Dispatch(client_dispatch)
 
-                    self._dispatches[dispatch.id] = dispatch
-                    old_dispatch = old_dispatches.pop(dispatch.id, None)
-                    if not old_dispatch:
+                    new_dispatches[dispatch.id] = dispatch
+                    old_dispatch = self._dispatches.get(dispatch.id, None)
+                    if old_dispatch is None:
                         _logger.debug("New dispatch: %s", dispatch)
                         await self._update_dispatch_schedule_and_notify(
                             dispatch, None, timer
@@ -396,23 +395,40 @@ class DispatchScheduler(BackgroundService):
                         )
                         await self._lifecycle_events_tx.send(Updated(dispatch=dispatch))
 
-            _logger.debug("Received %s dispatches", len(self._dispatches))
+            _logger.debug("Received %s dispatches", len(new_dispatches))
 
         except grpc.aio.AioRpcError as error:
             _logger.error("Error fetching dispatches: %s", error)
-            self._dispatches = old_dispatches
             return
 
-        for dispatch in old_dispatches.values():
+        # We make a copy because we mutate self._dispatches.keys() inside the loop
+        for dispatch_id in frozenset(self._dispatches.keys() - new_dispatches.keys()):
+            # Use try/except as the `self._dispatches` cache can be mutated by
+            # stream delete events while we're iterating
+            try:
+                dispatch = self._dispatches.pop(dispatch_id)
+            except KeyError as error:
+                _logger.warning(
+                    "Inconsistency in cache detected. "
+                    + "Tried to delete non-existing dispatch %s (%s)",
+                    dispatch_id,
+                    error,
+                )
+                continue
+
             _logger.debug("Deleted dispatch: %s", dispatch)
-            await self._lifecycle_events_tx.send(Deleted(dispatch=dispatch))
             await self._update_dispatch_schedule_and_notify(None, dispatch, timer)
 
-            # Set deleted only here as it influences the result of dispatch.started
-            # which is used in above in _running_state_change
+            # Set deleted only here as it influences the result of
+            # dispatch.started, which is used in
+            # _update_dispatch_schedule_and_notify above.
             dispatch._set_deleted()  # pylint: disable=protected-access
             await self._lifecycle_events_tx.send(Deleted(dispatch=dispatch))
 
+        # Update the dispatch list with the dispatches
+        self._dispatches.update(new_dispatches)
+
+        # Set event to indicate fetch ran at least once
         self._initial_fetch_event.set()
 
     async def _update_dispatch_schedule_and_notify(
--- a/src/frequenz/dispatch/_merge_strategies.py
+++ b/src/frequenz/dispatch/_merge_strategies.py
@@ -5,6 +5,7 @@
 
 import logging
 from collections.abc import Mapping
+from datetime import datetime, timezone
 from sys import maxsize
 from typing import Any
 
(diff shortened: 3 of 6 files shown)


ela-kotulska-frequenz (dismissed):

src/frequenz/dispatch/_merge_strategies.py
```diff
@@ -58,7 +58,8 @@ def filter(
         )
 
         _logger.debug(
-            "stop event %s because other_dispatches_running=%s",
+            "%s stop event %s because other_dispatches_running=%s",
+            "Ignoring" if other_dispatches_running else "Allowing",
             dispatch.id,
             other_dispatches_running,
         )
```
ela-kotulska-frequenz: You could print IDs of running dispatches ( if there are any )

pyproject.toml
```diff
@@ -40,7 +40,7 @@ dependencies = [
   # plugins.mkdocstrings.handlers.python.import)
   "frequenz-sdk >= 1.0.0-rc2100, < 1.0.0-rc2200",
   "frequenz-channels >= 1.6.1, < 2.0.0",
-  "frequenz-client-dispatch >= 0.11.0, < 0.12.0",
+  "frequenz-client-dispatch >= 0.11.1, < 0.12.0",
```
llucax, line 43: This dependency update seems to be misplaced, or is it necessary to call `super().started_at()`? If that is the case please mention this in the commit message, other we need to just guess when reviewing.

llucax (commented):
As usual, the PR and commit title and description could use some improvement. It is hard to review when they say so little. Could you please please please put a bit more thought into those? Please :pray: ? For the PR description just copying the copilot review summary is a big improvement.

For example in this PR it is mostly irrelevant that you are using `started_at`, that tells me almost nothing. What is important is you are fixing a bug caused by not having a consistent `now` used in different path of the code to decide on how to merge dispatches.

pyproject.toml
Marenz: Yeah, started_at was introduced in 0.11.1

Marenz:
>  commit title and description 

tbh, I disagree. The commit titles and descriptions are all perfectly fine imo.

PR, yeah, okay, that grew a bit and needed updating.

I also don't think a required dependency update necessarily needs to be mentioned if it is so minor as here, and the fact that its part of that particular commit is in itself a documentation of its need, at least that was my perspective.... 

I don't really see how _any_ of the other commit titles & content is lacking anything to make them understandable.

Marenz:
* Updated to print all other active dispatch ids
* Updated to mention the need of the version update in the according PR description.
* Updated another commit description to be even more clear

src/frequenz/dispatch/_merge_strategies.py
```diff
@@ -39,24 +42,38 @@ def filter(
...
+
+        if other_dispatches_running:
+            _logger.debug(
+                "Active other dispatches: %s",
+                (dispatch.id for dispatch in running_dispatch_list),
```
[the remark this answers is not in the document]
Marenz: oh damn.

llucax:
Without wanting to make a huge thing about this, I will just comment on this one, as I think it can bring some clarity:

> I also don't think a required dependency update necessarily needs to be mentioned if it is so minor as here, and the fact that its part of that particular commit is in itself a documentation of its need, at least that was my perspective....

I've seen many times people that sneak an unrelated change into a commit just because it was there. And I don't even think that's obscenely wrong, I can live with that when there is intention. The problem is when it was a mistake, and you were working on something else that needed the bump, but this PR doesn't really need it, and that dependency bump can have real world impact.

So what it is important for me is declaring intention. If the bump was needed by the commit, that's good to know, and good to know why, even if it could be inferred from the diff, just as an extra verification. If the bump was just done because one is lazy about putting it in a separate commit, **and the bump is harmless**, that should also be explicitly stated, so I can actually do a review, which is about detecting issues early.

This is my main point. I want to be able to tell between mistakes and intended changes.

src/frequenz/dispatch/_bg_service.py
```diff
@@ -396,23 +396,30 @@ async def _fetch(self, timer: Timer) -> None:
...
-            self._dispatches = old_dispatches
             return
 
-        for dispatch in old_dispatches.values():
+        # Delete old dispatches
+        for dispatch_id in old_dispatches:
+            if dispatch_id in new_dispatches:
+                continue
+
+            dispatch = self._dispatches.pop(dispatch_id)
```
llucax: `pop()` will raise if the key is not found, not sure if this could happen if you are dealing with some dict inconsistencies here. If it could happen you should either first check if it exists or catch the `KeyError` `pop` could raise.

src/frequenz/dispatch/_bg_service.py
```diff
@@ -396,23 +396,30 @@ async def _fetch(self, timer: Timer) -> None:
...
         except grpc.aio.AioRpcError as error:
             _logger.error("Error fetching dispatches: %s", error)
-            self._dispatches = old_dispatches
             return
 
-        for dispatch in old_dispatches.values():
+        # Delete old dispatches
+        for dispatch_id in old_dispatches:
+            if dispatch_id in new_dispatches:
+                continue
```
llucax: You could probably also do something like:
```suggestion
        for dispatch_id in old_dispatches.keys() - new_dispatches.keys():
```

llucax (dismissed):
LGTM, but one thing to check and one small suggestion.

src/frequenz/dispatch/_bg_service.py
Marenz: `old_dispatches == keys of self._dispatches`, so it is pretty safe, but I suppose there could come in a change during fetch from another part using stream I suppose..

will make it more fail safe

src/frequenz/dispatch/_bg_service.py
```diff
@@ -381,9 +381,9 @@ async def _fetch(self, timer: Timer) -> None:
                         continue
                     dispatch = Dispatch(client_dispatch)
 
-                    self._dispatches[dispatch.id] = dispatch
-                    old_dispatch = old_dispatches.pop(dispatch.id, None)
-                    if not old_dispatch:
+                    new_dispatches[dispatch.id] = dispatch
+                    old_dispatch = self._dispatches.get(dispatch.id, None)
```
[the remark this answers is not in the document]
Marenz: It's nice for the reader to see this explicitly I think

src/frequenz/dispatch/_bg_service.py
```diff
@@ -366,8 +366,8 @@ async def _fetch(self, timer: Timer) -> None:
         """
         self._initial_fetch_event.clear()
 
-        old_dispatches = self._dispatches
-        self._dispatches = {}
+        old_dispatches = set(self._dispatches.keys())
```
```suggestion
```

src/frequenz/dispatch/_bg_service.py
```diff
@@ -396,23 +396,33 @@ async def _fetch(self, timer: Timer) -> None:
...
+
+                # Set deleted only here as it influences the result of dispatch.started
+                # which is used in the func call above ^
+                dispatch._set_deleted()  # pylint: disable=protected-access
+                await self._lifecycle_events_tx.send(Deleted(dispatch=dispatch))
+            except KeyError:
+                _logger.warning(
+                    "Inconsistency in cache detected. Tried to delete non-existing dispatch %s",
+                    dispatch_id,
+                )
```
llucax: It is usually recommended to narrow the try/except to only the code you expect to fail, otherwise you could catch other failures that maybe are not OK to ignore. For example if some code in `await self._lifecycle_events_tx.send(Deleted(dispatch=dispatch))` (suppose `send()` uses some internal cache and there is a bug that also make it raise a `KeyError`).

Also if `old_dispatches` == `self._dispatches.keys()`, maybe it makes more sense to use `self._dispatches.keys()` directly now, I see in the new code `old_dispatches` is only being used here. This would make the code more obvious and actually guarantee the `pop()` can't fail, `old_dispatches` and `self._dispatches` can't get out of sync.

So I suggest removing `old_dispatches` and just do:

```suggestion
        # We make a copy because we mutate self._dispatch.keys() inside the loop
        for dispatch_id in frozenset(self._dispatches.keys() - new_dispatches.keys()):
            dispatch = self._dispatches.pop(dispatch_id)
            _logger.debug("Deleted dispatch: %s", dispatch)
            await self._update_dispatch_schedule_and_notify(None, dispatch, timer)

            # Set deleted only here as it influences the result of dispatch.started
            # which is used in the func call above ^
            dispatch._set_deleted()  # pylint: disable=protected-access
                await self._lifecycle_events_tx.send(Deleted(dispatch=dispatch))
```

llucax (commented):
LGTM, just one more suggestion, removing auto-merge in case you want to change it.

llucax (dismissed):

src/frequenz/dispatch/_bg_service.py
llucax: Not sure if the copy is necessary, as we are using the `-` which I guess produces a copy in itself, but I'm not 100% sure. Also making it a `frozenset` guarantees our items being iterated will never change, so it makes everything more explicit.

Marenz: > and actually guarantee the pop() can't fail, 

I don't think that's actually true. It can still fail, the maps can be modified between any call to await.

src/frequenz/dispatch/_merge_strategies.py
```diff
@@ -39,24 +42,38 @@ def filter(
...
             if (
                 self.identity(existing_dispatch) == self.identity(dispatch)
                 and existing_dispatch.id != dispatch.id
             )
         )
 
-        logging.debug(
-            "stop event %s because other_dispatches_running=%s",
+        other_dispatches_running = any(
+            dispatch.started_at(now) for dispatch in running_dispatch_list
```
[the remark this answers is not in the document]
Marenz: Slightly surprised pylint or flake didn't catch that

src/frequenz/dispatch/_bg_service.py
```diff
@@ -396,23 +395,38 @@ async def _fetch(self, timer: Timer) -> None:
...
+                dispatch = self._dispatches.pop(dispatch_id)
+            except KeyError as error:
+                _logger.warning(
+                    "Inconsistency in cache detected. "
+                    + "Tried to delete non-existing dispatch %s (%s)",
+                    dispatch_id,
+                    error,
+                )
+                continue
+
```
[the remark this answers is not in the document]
Marenz: shush

shsms (dismissed):

src/frequenz/dispatch/_bg_service.py
llucax: Oh, OK ,yeah, then maybe it is worth a comment in the `try`, as it might not be obvious to the casual reader.

llucax (dismissed):
LGTM too, one small optional comment about maybe adding a comment to the `try` for `pop()` as it might not be obvious.

PR #184, opened 2025-08-20
Support key signing and handle dry_run cleaner

### LINKED ISSUE #132: Prevent dry_run dispatch from overriding live dispatch
Based on the discussion, it seems that a dry_run dispatch can potentially override a live dispatch, which could lead to inconsistencies. The following points were raised:

An actor running a dispatch should expect the dry_run state to change during the run, which can cause inconsistencies.
It was suggested to prevent this behavior to maintain consistency with the update behavior.

One approach could be to split the groups into dry and wet run dispatches, allowing them to run in parallel without interference.

The main use case for dry_run is for testing purposes, where it should not interrupt the main actor.
Supporting parallel running of dry_run and live dispatches would allow users to test new values without affecting the live dispatch.

Marenz:
- **Support key signing**
- **Consider dry_run status in dispatch merging**

### DIFF
--- a/src/frequenz/dispatch/_dispatcher.py
+++ b/src/frequenz/dispatch/_dispatcher.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import warnings
 from asyncio import Event
 from datetime import timedelta
 from typing import Awaitable, Callable, Self
@@ -64,7 +65,7 @@ class Dispatcher(BackgroundService):
             async with Dispatcher(
                 microgrid_id=microgrid_id,
                 server_url=url,
-                key=key
+                auth_key=key
             ) as dispatcher:
                 dispatcher.start_managing(
                     dispatch_type="DISPATCH_TYPE",
@@ -90,7 +91,7 @@ class Dispatcher(BackgroundService):
             async with Dispatcher(
                 microgrid_id=microgrid_id,
                 server_url=url,
-                key=key
+                auth_key=key
             ) as dispatcher:
                 actor = MagicMock() # replace with your actor
 
@@ -135,7 +136,7 @@ class Dispatcher(BackgroundService):
             async with Dispatcher(
                 microgrid_id=microgrid_id,
                 server_url=url,
-                key=key,
+                auth_key=key,
             ) as dispatcher:
                 events_receiver = dispatcher.new_lifecycle_events_receiver("DISPATCH_TYPE")
 
@@ -172,7 +173,7 @@ class Dispatcher(BackgroundService):
             async with Dispatcher(
                 microgrid_id=microgrid_id,
                 server_url=url,
-                key=key,
+                auth_key=key,
             ) as dispatcher:
                 # Create a new dispatch
                 new_dispatch = await dispatcher.client.create(
@@ -205,7 +206,9 @@ class Dispatcher(BackgroundService):
         *,
         microgrid_id: MicrogridId,
         server_url: str,
-        key: str,
+        key: str | None = None,
+        auth_key: str | None = None,
+        sign_secret: str | None = None,
         call_timeout: timedelta = timedelta(seconds=60),
         stream_timeout: timedelta = timedelta(minutes=5),
     ):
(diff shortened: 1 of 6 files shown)


tests/test_managing_actor.py
```diff
@@ -335,20 +335,25 @@ async def test_manage_abstraction(
...
-            key: str,
+            auth_key: str | None = None,
+            key: str | None = None,
+            sign_secret: str | None = None,
             call_timeout: timedelta,
             stream_timeout: timedelta,
         ) -> None:
             assert server_url
-            assert key
+            assert key is None or auth_key is None
```
llucax: And
```suggestion
            assert key or auth_key
```

?

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -214,15 +216,29 @@ def __init__(
...
+            key: The key to access the service, deprecated, use `auth_key` instead.
+            auth_key: The authentication key to access the service.
+            sign_secret: The secret to sign the requests, optional
             call_timeout: The timeout for API calls.
             stream_timeout: The timeout for streaming API calls.
+
+        Raises:
+            ValueError: If both `key` and `auth_key` are provided.
         """
         super().__init__()
```
llucax, line 229: You could add a:

```py
if key:
    warnings.warn("key is deprecated, use auth_key instead", DeprecationWarning, stack_whatever=2)
```

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -214,15 +216,29 @@ def __init__(
...
 
+        if key is not None and auth_key is not None:
+            raise ValueError(
+                "Both 'key' and 'auth_key' are provided, please use only 'auth_key'."
+            )
+
+        if key is not None:
+            auth_key = key
```
llucax, line 242: It would be silly to do, but maybe also add an exception for `if not key and not auth_key`? Or will this be caught by `DispatcherApiClient()`?

llucax (commented):
Awesome to see the support for parallel dry-run actors! 🎉 

Next time maybe split the deprecation of `key` in favor of `auth_key` in a separate commit, as it was sometimes confusing to have it mixed with the introduction of signing.

tests/test_managing_actor.py
Marenz: Really? The one time I am explict about the type checking :laughing:

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -214,15 +217,37 @@ def __init__(
...
         super().__init__()
 
+        if key is not None and auth_key is not None:
+            raise ValueError(
+                "Both 'key' and 'auth_key' are provided, please use only 'auth_key'."
+            )
+
+        if key is None and auth_key is None:
+            raise ValueError(
+                "Either 'key' or 'auth_key' must be provided to access the dispatch service."
```
llucax: Niptpick: Since `key` is deprecated I wouldn't mention it here.
```suggestion
                "'auth_key' must be provided to access the dispatch service."
```

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -214,15 +217,37 @@ def __init__(
...
+            )
```
llucax: It just occurred to me, but you could also use overloads to ensure this is called correctly when type-checking:

```py
    @overload
    def __init__(
        self,
        *,
        microgrid_id: MicrogridId,
        server_url: str,
        key: str,
        auth_key: None = None,
        sign_secret: str | None = None,
        call_timeout: timedelta = timedelta(seconds=60),
        stream_timeout: timedelta = timedelta(minutes=5),
    ):
    @overload
    def __init__(
        self,
        *,
        microgrid_id: MicrogridId,
        server_url: str,
        key: None = None,
        auth_key: str,
        sign_secret: str | None = None,
        call_timeout: timedelta = timedelta(seconds=60),
        stream_timeout: timedelta = timedelta(minutes=5),
    ):
```

But might be overkill for this.

src/frequenz/dispatch/_dispatcher.py
```diff
@@ -214,15 +217,37 @@ def __init__(
...
+
+        if key is not None:
+            auth_key = key
+            warnings.warn(
+                "'key' is deprecated, use 'auth_key' instead.", DeprecationWarning
```
llucax: Add `stacklevel=2` so the deprecation is shown in the call site instead of here?

llucax: I suggested this before copilot, damn coffee spilling that interrupted my review! ☕ ⌨️

llucax (dismissed):
Just a couple of nitpick comments.

PR #195, opened 2025-09-29
Fetch only ongoing dispatches

Marenz:
Based on https://github.com/frequenz-floss/frequenz-client-dispatch-python/pull/212

### DIFF
--- a/tests/test_frequenz_dispatch.py
+++ b/tests/test_frequenz_dispatch.py
@@ -476,9 +476,7 @@ async def test_dispatch_new_but_finished(
     )
     fake_time.shift(timedelta(seconds=1))
 
-    await asyncio.sleep(1)
-    # Process the lifecycle event caused by the old dispatch at startup
-    await test_env.lifecycle_events.receive()
+    # Expecting no event for a past and finished dispatch as it should be filtered out
 
     # Create another dispatch the normal way
     new_dispatch = generator.generate_dispatch()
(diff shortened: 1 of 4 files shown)


tests/test_frequenz_dispatch.py
```diff
@@ -524,23 +523,31 @@ async def test_notification_on_actor_start(
...
 
-    fake_time.shift(timedelta(seconds=1))
-    await asyncio.sleep(1)
+    fake_time.shift(timedelta(seconds=7))
+    await asyncio.sleep(6)
+    logging.warning("Time shifted")
 
     # Expect notification of the running dispatch being ready to run
     ready_dispatch = await test_env.running_state_change.receive()
+    logging.warning("Received running state change: %s", ready_dispatch)
```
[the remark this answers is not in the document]
Marenz: Removed logging

tests/test_frequenz_dispatch.py
```diff
@@ -476,9 +476,7 @@ async def test_dispatch_new_but_finished(
     )
     fake_time.shift(timedelta(seconds=1))
 
-    await asyncio.sleep(1)
-    # Process the lifecycle event caused by the old dispatch at startup
-    await test_env.lifecycle_events.receive()
+    # Expecting no event for a past and finished dispatch as it should be filtered out
```
llucax, line 479: What about the 5 seconds in the past? Wouldn't it be possible to get already finished dispatches in some edge cases?

llucax (commented):
Nice, LGTM, no approving only because of the PR dependency.

tests/test_frequenz_dispatch.py
Marenz: Yes, but that's not an issue, we still manage/filter them on our side, the server-filter is just an optimization

Marenz: And we rather have too much than too little, thus the 5 second buffer

llucax (approved):

PR #240, opened 2026-02-10
Use common API credential variables in examples

Marenz:
Updates credential examples to use `FREQUENZ_API_KEY` and `FREQUENZ_API_SECRET`, following frequenz-io/rnd#187.

No runtime credential handling changes.


llucax:
I would hold on this, but it looks like we are moving towards removing the `auth_` and `secret_` prefixes, so I would be fine to merge this if you are extra eager. But we need to build consensus and change all other libraries, otherwise it just creates confusion.

PR #246, opened 2026-03-11
Fix scheduler bugs: heap invariant corruption and lost stop event at dispatch boundary

Marenz:
This PR fixes two scheduler bugs when updating an existing dispatch.

1. Removing a queued event could corrupt the scheduler heap because the queue was not re-heapified afterward, which could lead to incorrect ordering of future events.
2. An update exactly at the stop boundary could remove the pending stop event without notifying the actor, causing the stop transition to be missed.

This PR restores the heap after removals, tracks which queued event was removed, and sends the missing stop notification when a boundary update unschedules the pending stop event.

Regression tests and release notes were added.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -20,3 +20,8 @@
 ## Bug Fixes
 
 <!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+
+* Fixed bugs that could cause a dispatch actor not to stop when an update
+  arrives exactly at the moment a dispatch window closes. Also fixed a
+  related issue that could cause incorrect event ordering when multiple
+  dispatches are updated concurrently.
--- a/src/frequenz/dispatch/_bg_service.py
+++ b/src/frequenz/dispatch/_bg_service.py
@@ -486,10 +486,18 @@ class DispatchScheduler(BackgroundService):
         # Dispatch was updated
         elif dispatch and old_dispatch:
             # Remove potentially existing scheduled event
-            self._remove_scheduled(old_dispatch)
+            removed = self._remove_scheduled(old_dispatch)
 
             # Check if the change requires an immediate notification
             if self._update_changed_running_state(dispatch, old_dispatch):
+                await self._send_running_state_change(dispatch)
+            elif removed is not None and removed.priority == 1 and not dispatch.started:
+                # priority == 1 means a stop event (see QueueItem.__init__).
+                # If we removed a pending stop event and the dispatch is no
+                # longer started, the update arrived exactly at the stop
+                # boundary. The timer would have delivered the stop event, but
+                # _remove_scheduled consumed it first. Send the notification
+                # here so the actor is not left running past the window end.
                 await self._send_running_state_change(dispatch)
 
             if dispatch.started:
--- a/src/frequenz/dispatch/_bg_service.py
+++ b/src/frequenz/dispatch/_bg_service.py
@@ -13,7 +13,7 @@ from collections.abc import Mapping
 from contextlib import closing
 from dataclasses import dataclass, field
 from datetime import datetime, timedelta, timezone
-from heapq import heappop, heappush
+from heapq import heapify, heappop, heappush
 
 import grpc.aio
 from frequenz.channels import Broadcast, Receiver, select, selected_from
@@ -507,21 +507,26 @@ class DispatchScheduler(BackgroundService):
             timer.reset(interval=due_at - datetime.now(timezone.utc))
             _logger.debug("Next event scheduled at %s", self._scheduled_events[0].time)
 
-    def _remove_scheduled(self, dispatch: Dispatch) -> bool:
+    def _remove_scheduled(self, dispatch: Dispatch) -> "QueueItem | None":
         """Remove a dispatch from the scheduled events.
 
         Args:
             dispatch: The dispatch to remove.
 
         Returns:
-            True if the dispatch was found and removed, False otherwise.
+            The removed queue item, or None if not found.
         """
         for idx, item in enumerate(self._scheduled_events):
             if dispatch.id == item.dispatch.id:
                 self._scheduled_events.pop(idx)
-                return True
+                # heappop() only removes the root (index 0) and does not accept
+                # an index argument, so we use list.pop(idx) instead. After
+                # removing an arbitrary element the heap property is broken and
+                # must be restored explicitly.
+                heapify(self._scheduled_events)
+                return item
 
-        return False
+        return None
 
     def _schedule_start(self, dispatch: Dispatch) -> None:
         """Schedule a dispatch to start.
--- a/tests/test_dispatcher_scheduler.py
+++ b/tests/test_dispatcher_scheduler.py
@@ -1,0 +1,302 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+# pylint: disable=too-many-arguments,too-many-positional-arguments,too-many-locals
+
+"""Tests for scheduler edge cases in DispatchScheduler."""
+
+import asyncio
+from collections.abc import AsyncIterator, Iterator
+from dataclasses import replace
+from datetime import datetime, timedelta, timezone
+from random import randint
+
+import async_solipsism
+import time_machine
+from frequenz.channels import Receiver
+from frequenz.client.common.microgrid import MicrogridId
+from frequenz.client.dispatch.recurrence import RecurrenceRule
+from frequenz.client.dispatch.test.client import FakeClient, to_create_params
+from frequenz.client.dispatch.test.generator import DispatchGenerator
+from pytest import fixture
+
+from frequenz.dispatch import Dispatch, DispatchEvent
+from frequenz.dispatch._bg_service import DispatchScheduler
+
+
+@fixture
+def generator() -> DispatchGenerator:
+    """Return a dispatch generator."""
+    return DispatchGenerator()
+
+
+@fixture
+def event_loop_policy() -> async_solipsism.EventLoopPolicy:
+    """Set the event loop policy to use async_solipsism."""
+    policy = async_solipsism.EventLoopPolicy()
+    asyncio.set_event_loop_policy(policy)
+    return policy
+
+
+@fixture
+def fake_time() -> Iterator[time_machine.Coordinates]:
+    """Replace real time with a time machine that doesn't automatically tick."""
+    with time_machine.travel(destination=0, tick=False) as traveller:
+        yield traveller
+
+
+def _now() -> datetime:
+    """Return the current time in UTC."""
+    return datetime.now(tz=timezone.utc)
+
+
+@fixture
+def microgrid_id() -> MicrogridId:
+    """Return a random microgrid ID."""
+    return MicrogridId(randint(1, 100))
+
+
+@fixture
+def client() -> FakeClient:
+    """Return a fake dispatch API client."""
+    return FakeClient()
+
+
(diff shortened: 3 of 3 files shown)


src/frequenz/dispatch/_bg_service.py
```diff
@@ -507,21 +513,22 @@ def _update_timer(self, timer: Timer) -> None:
...
         Returns:
-            True if the dispatch was found and removed, False otherwise.
+            The removed queue item, or None if not found.
         """
         for idx, item in enumerate(self._scheduled_events):
             if dispatch.id == item.dispatch.id:
                 self._scheduled_events.pop(idx)
-                return True
+                heapify(self._scheduled_events)
+                return item
```
llucax, line 535: I guess you are not using `heappop()` instead of `.pop()` + `heapify()` here because you are extracting an element with an arbitrary priority, right? Maybe it is worth a comment in case someone tries to "fix it" in the future?

src/frequenz/dispatch/_bg_service.py
```diff
@@ -486,11 +486,17 @@ async def _update_dispatch_schedule_and_notify(
         # Dispatch was updated
         elif dispatch and old_dispatch:
             # Remove potentially existing scheduled event
-            self._remove_scheduled(old_dispatch)
+            removed = self._remove_scheduled(old_dispatch)
 
             # Check if the change requires an immediate notification
             if self._update_changed_running_state(dispatch, old_dispatch):
                 await self._send_running_state_change(dispatch)
+            elif removed is not None and removed.priority == 1 and not dispatch.started:
```
llucax, line 494: It took me some amount of digging to figure out `priority == 1` basically means it is a "stop", is this correct?

I wonder if we can this information from some more clear place than the priority, and if not, if it could make sense to make the priority a `IntEnum` instead, so we can write `priority == Priority.STOP_EVENT`, but it also feels a bit weird that priority is "start" or "stop" as it is not clear without context which should have higher priority.

In any case, if we don't find a clearer way to express this, I would at least make it more clear in the comment that `priority == 1` means it is a "stop event".

src/frequenz/dispatch/_bg_service.py
```diff
@@ -486,11 +486,17 @@ async def _update_dispatch_schedule_and_notify(
...
+                # Before this fix, an update exactly at the stop boundary could
+                # remove the pending stop event and skip the notification. The
+                # dispatch window has elapsed, so the actor still needs the stop
+                # event that the timer would have sent.
```
llucax: This also sounds weird to me, it only makes sense in the context of the PR, once the PR is merged this is not a "fix" anymore, so I would phrase it differently, like "We need to check for this case because..." (and tied to the above comment, maybe say something like "at the stop boundary (priority == 1 means it is a stop event)".

RELEASE_NOTES.md
```diff
@@ -21,3 +21,7 @@ This is the first major version release with authentication parameter updates an
 ## Bug Fixes
 
 <!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+
+* Fixed two dispatch rescheduling issues for unmerged dispatches:
+  * restore the scheduler heap invariant after removing queued events during updates
+  * resend the missing stop notification when an update removes a pending stop event at the dispatch boundary
```
llucax: I think this misses the most important information, what the "issues" were. This should target library users, not developers. What users care about is what didn't work before and works now, not about what changed internally, this is not a changelog. I guess something like "Fixed bugs that could cause actors not to stop if an update event came exactly at the same time an actor should stop" should be enough.

tests/test_scheduler_regressions.py
```diff
@@ -0,0 +1,283 @@
...
+    This mirrors the SET_POWER edge case where multiple dispatches share the
+    exact same start time and duration, and another batch starts exactly when
+    the first batch ends.
+
+    This is an opportunistic coverage test: it also passed before the heap fix.
+    The broken heap ordering depends on runtime queue layouts that this test
+    harness does not reproduce reliably, but this still exercises the production
+    traffic shape around shared boundaries.
+    """
+    env = await _new_env()
```
llucax: This feels pretty weird, I was confused by the "env", I though it was something to do with env vars.

I don't remember how tests are shaped in this repo, but I strongly recommend using fixtures instead, we can reuse them for other tests and makes test dependency on some "environment" much more explicit and clear. Also an error with the "env" setup is reported as a "test error" instead of a "test failure".

```py
@fixture
def microgrid_id() -> MicrogridId:  # this can also be a global
    return MicrogridId(randint(1, 100))

@fixture
def client() -> DispatchApiClient:
    return FakeClient()

@fixture
async def scheduler(microgrid_id: MicrogridId, client: DispatchApiClient) -> AsyncIterator[DispatchScheduler]:
    sched = DispatchScheduler(microgrid_id=microgrid_id, client=client)
    sched.start()
    try:
        yield sched
    finally:
        await sched.stop()
    # or `async with sched: yield sched` if the scheduler is a context manager

    receiver = await service.new_running_state_event_receiver(
        "SET_POWER", merge_strategy=None
    )

@fixture
async def event_receiver(scheduler: DispatchScheduler) -> Receiver[?]:
    return await scheduler.new_running_state_event_receiver(
        "SET_POWER", merge_strategy=None
    )

@fixture
def lifecycle_receiver(scheduler: DispatchScheduler) -> Receiver[?]:
    return scheduler.new_lifecycle_events_receiver("SET_POWER")

async def test_parallel_dispatches_same_window_and_follow_up_window(
    fake_time: time_machine.Coordinates,
    generator: DispatchGenerator,
    client: DispatchClient,
    event_receiver: Receiver[?],
    lifecycle_receiver: Receiver[?],
) -> None:
```

tests/test_scheduler_regressions.py
```diff
@@ -0,0 +1,283 @@
...
+    """Test that unmerged dispatches stop correctly at a shared boundary.
+
+    This mirrors the SET_POWER edge case where multiple dispatches share the
+    exact same start time and duration, and another batch starts exactly when
+    the first batch ends.
+
+    This is an opportunistic coverage test: it also passed before the heap fix.
+    The broken heap ordering depends on runtime queue layouts that this test
+    harness does not reproduce reliably, but this still exercises the production
+    traffic shape around shared boundaries.
```
llucax: Again, talking about "the heap fix" will go out of context as soon as this PR is merged. In this case I guess you could say something more general, like "this test is opportunistic and tries to detect corruption in the _scheduled_events heap". An alternative would be to have an issue that describes "the heap bug" and then reference that issue for more context.

tests/test_dispatcher_scheduler.py
llucax, line 1: This file sounds weird, again it sounds good in the context of the PR, but once this is merged, this is just a test, all tests are supposed to detect regressions, right?

I see this was added in `test_frequenz_dispatcher.py` and then moved here:

> so the main scheduler test file stays below pylint's module-size threshold and no targeted lint exceptions are needed.

I don't think this is a reason enough, we can either ignore that or if we really think it is an issue having such big files, split the tests more consciously, maybe having a separate `test_dispatcher_scheduler.py` for scheduler tests.

If you think these tests are niche enough to be separated, I guess that could make sense too, but then I would create an issue and rename this to `test_scheduler_regressions_issue_xxxx.py` to make it explicit about a particular issue.

llucax (commented):
I have a few comments about structure mainly. Also a few extra minor comments:

1. No empty line between the last line and signed-off-by in the commit message:

    ```
    Fix: call heapify() after pop(idx) to restore the heap invariant.
    Signed-off-by: Mathias L. Baumann <mathias.baumann@frequenz.com>
    ```
2. The release notes update is mixed in the last commit, I guess it makes more sense to put it in its own commit.

3. The PR description also IMHO focus on the wrong stuff, it didn't help a lot understanding what was not working. Asking AI to create a better description that explains what was exactly the bug, and how it was fixed by the PR, it came up with:

    > This PR fixes two scheduler bugs when updating an existing dispatch.
    > 1. Removing a queued event could corrupt the scheduler heap because the queue was not re-heapified afterward, which could lead to incorrect ordering of future events.
    > 2. An update exactly at the stop boundary could remove the pending stop event without notifying the actor, causing the stop transition to be missed.
    > 
    > This PR restores the heap after removals, tracks which queued event was removed, and sends the missing stop notification when a boundary update unschedules the pending stop event.
    > 
    > Regression tests and release notes were added.

PS: TIL: re-sifted == re-heapified

llucax:
One more question. Could the boundary issue happen for updates happen at the **start** boundary instead of the **stop** boundary?

src/frequenz/dispatch/_bg_service.py
Marenz: You cannot use `heappop()` because it does not take parameters for the required `idx`, so any "fix it" attempt would fail at the function definition. I can add a comment though.

llucax: Right. If it can't be really replaced by `heappop()` I guess a comment is not really necessary, but if it is very low friction I guess it could help.

I wonder if adding an OO-abstraction over `heap*()` could help (maybe in `core`) avoiding this kind of bugs. It feels very fragile to use `heap*()` directly over a list, very easy to mess with the underlying storage of a heap accidentally. For this case we would need to provide a way to get an element an arbitrary index, by keeping the heap a heap, but I think it is a valid use of the heap, as long as we say it is an expensive operation.

Marenz:
Okay, I think I addressed all feedback now

Marenz:
>One more question. Could the boundary issue happen for updates happen at the start boundary instead of the stop boundary?

The lost-event bug only occurs at the stop boundary, not the start boundary.

At the start boundary, the update path already has a fallback notification path through `_update_changed_running_state()`. The key point is that `started` is one of the `runtime_state_attributes`, and `started` is derived from the dispatch window and the current time, not stored independently.

That means:
- just before the start boundary, `old_dispatch.started == False`
- once the boundary is crossed, `dispatch.started == True`
- 
So even if `_remove_scheduled()` consumed the queued start event first, `_update_changed_running_state()` would still see the `started` transition and emit the running-state update.

At the stop boundary, the problematic case is different:

- `_remove_scheduled()` can consume the queued stop event
- but the concurrent update may not independently trigger `_update_changed_running_state()`
In that case, nothing replaces the consumed stop event, so the stop notification is lost.

src/frequenz/dispatch/_bg_service.py
Marenz: Yeah, a proper datastructure surely would have prevented this particular one, true.

llucax: * https://github.com/frequenz-floss/frequenz-core-python/issues/143

llucax (approved):
Awesome!

llucax:
Wow, no merge queue for this repo?

llucax:
On a second thought, I'm doubtful the merge queue really adds any value to us, I guess the try-merge and remove, go to the next in the queue is good for monorepos, that might have many many concurrent merges, but for us, that we hardly have any (maybe only when dependabot kicks in), it is starting to feel it just makes things slower and more expensive (testing twice), without much extra gain. But I guess this is something to discuss elsewhere...

Marenz:
Shit. I targeted the wrong upstream branch. We are on `v1.x.x`

Marenz:
And that's also why there is no merge queue.

llucax:
Maybe remove the v0.x.x branch if we are not making more releases from it?

Marenz:
> Maybe remove the v0.x.x branch if we are not making more releases from it?

already done :P

PR #248, opened 2026-03-12
Fix scheduler bugs on v1.x.x

Marenz:
Rebased follow-up of #246 for `v1.x.x`.

- restores the scheduler heap after removing queued events
- emits the missing stop notification when an update lands on the stop boundary
- adds regression coverage for both scheduler issues
- adapts the new scheduler tests to the `time-machine` 3 API used on `v1.x.x`


Marenz:
This is essentially the same as the already approved PR, just on the correct branch. Merging..

PR #279, opened 2026-06-30
Fix and improve docstring documentation

llucax:
🤖 This pull request is part of an automated effort to improve the docstring documentation of all our projects.

Main issues being resolved:
- Typos: "highlevel" → "high-level" throughout
- Class docstring summaries missing required "A"/"An" article
- Boolean method summaries using "Check if ..." instead of "Return whether ..."
- Property docstrings using imperative mood instead of noun phrases
- Missing \`Args:\`, \`Returns:\`, and \`Yields:\` sections in several methods
- Missing inline attribute docstrings on \`QueueItem\` dataclass fields
- Bare backtick symbol mentions that should be cross-reference links
- Absolute cross-references replaced with relative ones where applicable
- Types appearing inside docstring sections (Yields: datetime:)
- Spurious \`Returns:\` section on a generator method (\`missed_runs\`)
- Redundant sentence repeating the summary in \`missed_runs\`
- Non-standard \`Deprecation:\` admonition replaced with \`Warning:\`
- Missing imports and \`await\` in fenced code examples
- Deprecated \`.components\` usage replaced with \`.target\` in example
- Factually incorrect module docstring in \`conftest.py\` (claimed pylint, is Sybil)
- Stale comment describing \`QueueItem\` as a tuple (it is a dataclass)

Other things to note:
- The \`relative_crossrefs: true\` mkdocstrings option was added to \`mkdocs.yml\` as part of the rolling-out of relative cross-reference conventions.
- Three new pydoclint options were added to \`pyproject.toml\` (\`check-class-attributes\`, \`check-style-mismatch\`, \`require-inline-class-var-docs\`) as part of the rolling-out of stricter docstring checks.
- The \`ActorDispatcher\` example code block was re-indented to sit properly inside the \`Example:\` admonition, which also required adding missing imports (\`Any\`, \`Self\`) and an \`await\` on \`new_running_state_event_receiver\`.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -12,6 +12,10 @@
 
 <!-- Here goes the main new features and examples or instructions on how to use them -->
 
+## Enhancements
+
+- Improved docstring documentation across the project.
+
 ## Bug Fixes
 
 <!-- Here goes notable bug fixes that are worth a special mention or explanation -->
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -112,6 +112,7 @@ plugins:
             show_source: true
             show_symbol_type_toc: true
             signature_crossrefs: true
+            relative_crossrefs: true
           inventories:
             # See https://mkdocstrings.github.io/python/usage/#import for details
             - https://docs.python.org/3/objects.inv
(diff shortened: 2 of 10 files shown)


Marenz (approved):

Marenz:
Manually reviewed!!!

